PR #6, opened 2019-04-17
remove conf dir and various updates

siyopao:
Remove the conf directory, update the commands to look for client.properties in the project directory, and update the readme


explorer/README.md
```diff
@@ -14,9 +14,9 @@ After a successful run the `explorer` executable may be found in `/build/install
 
 ## How to use the command-line interface
 
-The first thing you will need to do is to edit the [client.properties](./conf/client.properties) file (found in the `conf` directory) to suit your setup. `client.properties` is a configuration of Scalar server's and user's information. Scalar DL Explorer uses it to connect to a specified Scalar server and keep track of the user's private key.
+The first thing you will need to do is to create a [`client.properties`](./client.properties) file. The `client.properties` is used to specify the configuration of a Scalar DL server and the user's information. If you have already used Scalar DL then you probably already have a `client.properties` file. Simply copy it to the project directory.
```
feeblefakie: Maybe, copying the file is not a good idea because if you update one you need to update another.
Can we just say, use your properties file simply ?

explorer/client.properties
```diff
@@ -1,19 +1,18 @@
 # Optional. A hostname of the Server. Use localhost by default if not specified.
-scalar.ledger.client.server_host=localhost
```
feeblefakie: I think we can remove this because a user surely has done the execution with his client.properties and has his own.

explorer/README.md
siyopao: Thank you. Done!

explorer/client.properties
siyopao: Deleted!

PR #18, opened 2020-03-01
Handle spaces and user inputs more properly

### LINKED ISSUE #14: JSON with spaces make the emulator crash
In the emulator, below command will return an error...
```
put "te st" {"state": 1}
execute get {" asset_id": "X"}
```
below will not be registered as a command...
```
get "te st"
scan "te st"
```

The desired output is where spaces can be considered as input.
Also it should works as expected for UDF as well. Such that `execute put {"asset_id": "foo", "data": {"alice": 70, "bob": 320}, "_functions_":["state-updater-function"]} -fa={"asset_id ": "bar"}` should works

**Proposed Solution**
Create a generic parser for user input (ex: `"te st"`) so that it can parse it to be a string as a whole instead of separated `"te"` and `"st"`.

scalarindetail:
This PR handles fixes #14 ...


emulator/src/main/java/com/scalar/client/tool/emulator/EmulatorTerminal.java
```diff
@@ -211,4 +215,20 @@ private boolean parseAndRunCommand(String line) {
     }
     return false;
   }
+
+  private String[] getParamsArray(String params, String commandName) {
+    if (commandName.matches("register[\\w-]*")) {
+      return params.split(" ");
+    }
+    List<String> allMatches = new ArrayList<String>();
+    Pattern pattern = Pattern.compile("(\".+?\"|[\\w.-]*)(\\s.+|.*)");
```
scalarindetail: This pattern can be better explain [here](https://regex101.com/r/7CxhXb/2/tests)...

emulator/src/main/java/com/scalar/client/tool/emulator/EmulatorTerminal.java
```diff
@@ -211,4 +214,33 @@ private boolean parseAndRunCommand(String line) {
...
+        }
+      }
+
+      byteArray.write(params.charAt(i));
+      if ((character.matches(" ") && stack.empty()) || i == params.length() - 1) {
+        paramsList.add(byteArray.toString().trim());
+        byteArray.reset();
+      }
+    }
+    return paramsList.toArray(new String[0]);
```
scalarindetail: Basically this is a stack and queues implementation...
1. Each character of input will be streamed via `param.charAt(i)`
2. Whenever a special character (`"` or `{` or `}`) is detected, it will be pushed into the stack.
3. ONLY pop IF `"` detected and last pushed character is `"`. OR IF `}` detected and last pushed character `{`.
4 `byteArray` acts as a buffer, it will take in characters until `(character.matches(" ") && stack.empty()) || i == params.length() - 1)` condition is met...if so, its content will be added to the paramsList and then `.reset()`.

emulator/src/main/java/com/scalar/client/tool/emulator/EmulatorTerminal.java
```diff
@@ -211,4 +214,33 @@ private boolean parseAndRunCommand(String line) {
...
+        }
+      }
+
+      byteArray.write(params.charAt(i));
+      if ((character.matches(" ") && stack.empty()) || i == params.length() - 1) {
+        paramsList.add(byteArray.toString().trim().replaceAll("^\"|\"$", ""));
+        byteArray.reset();
+      }
+    }
+    return paramsList.toArray(new String[0]);
```
scalarindetail: Basically this is a stack and queues implementation...
1. Each character of input will be streamed via `param.charAt(i)`
2. Whenever a special character (`"` or `{` or `}`) is detected, it will be pushed into the stack.
3. ONLY pop IF `"` detected and last pushed character is `"`. OR IF } detected and last pushed character {.
4. `byteArray` acts as a buffer, it will take in characters until `(character.matches(" ") && stack.empty()) || i == params.length() - 1)` condition is met...if so, its content will be added to the `paramsList` and then `.reset()`.

Torch3333: Nice, thanks for the detailed explanation. 
Since the algorithm is quite complicated, maybe we should write a few unit tests for it.

PR #19, opened 2020-03-02
Create and setup CI for Scalar DL Emulator with Github Actions

scalar-boney:
Task: https://scalar-labs.atlassian.net/browse/DLT-5310

### DIFF
--- a/.github/workflows/emulator.yml
+++ b/.github/workflows/emulator.yml
@@ -19,7 +19,7 @@ jobs:
         java-version: 1.8
 
     - name: Build Emulator
-      run: |
-        cd emulator
-        ./gradlew build
-        ./gradlew installDist
+      run: ./gradlew assemble
+
+    - name: Test Emulator
+      run: ./gradlew test
--- a/.github/workflows/emulator.yml
+++ b/.github/workflows/emulator.yml
@@ -1,0 +1,19 @@
+name: Emulator-CI
+
+on: [push]
+
+jobs:
+  Emulator:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    - name: Build with Gradle
+      run: ./gradlew build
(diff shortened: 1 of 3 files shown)


.github/workflows/emulator.yml
```diff
@@ -0,0 +1,22 @@
+name: Emulator-CI
+
+on: [push]
```
yito88: Is this executed when any push happens for other directories?
For example, when the explorer is updated, this CI for the emulator will be executed.
I think it is better to use `paths`.
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

scalar-boney: Thank you for review comment.
Workflow modified based on the above comment.

.github/workflows/emulator.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+    - name: Clone Scalardl-tool repository
+      uses: actions/checkout@v2
+
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
+    - name: Build Emulator
```
feeblefakie: `build` will trigger tests as well, so why don't you separate the runs like below ?
```
- name: Build Emulator
  run: ./gradlew assemble

- name: Test Emulator
  run: ./gradlew test
```

.github/workflows/emulator.yml
```diff
@@ -0,0 +1,25 @@
...
+  push:
+    paths:
+      - 'emulator/**'
+
+jobs:
+  Emulator:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Clone Scalardl-tool repository
```
feeblefakie:
```suggestion
    - name: Clone scalardl-tools repository
```
BTW do we need this ? seems obvious.

.github/workflows/emulator.yml
scalar-boney: Thank you review comment.
Workflow modified based on review comments.

.github/workflows/emulator.yml
scalar-boney: Thank you review comment.
Workflow modified based on review comments.

yito88: I think `jobs.<job_id>.steps.name` is usefule since GitHub Action displays that step name.
We can immediately know which step fails.

yito88:
@scalar-boney Why not use `actions/cache` to reduce CI time?
https://github.com/actions/cache

Issue #15, opened 2020-02-25
getCertificateKey() is not supported


scalarindetail:
The new method `getCertificateKey()` in Contract class in Scalar DL 2.0 is not supported.

PR #24, opened 2020-03-06
Add set-certificate command

Torch3333:
This PR resolves https://github.com/scalar-labs/scalardl-tools/issues/15.

### DIFF
--- a/emulator/src/main/java/com/scalar/client/tool/emulator/ContractManagerEmulator.java
+++ b/emulator/src/main/java/com/scalar/client/tool/emulator/ContractManagerEmulator.java
@@ -24,12 +24,14 @@ import javassist.CtNewMethod;
 import javassist.Modifier;
 import javax.json.Json;
 import javax.json.JsonObject;
+import org.apache.commons.lang3.reflect.FieldUtils;
 
 public class ContractManagerEmulator {
 
   private final ContractRegistry registry;
   private final Map<String, Contract> cache;
   public ContractManagerEmulator that;
+  private CertificateEntry.Key emulatedCertificateKey;
 
   public ContractManagerEmulator(ContractRegistry registry) {
     this.registry = registry;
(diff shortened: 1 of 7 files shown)


emulator/src/main/java/com/scalar/client/tool/emulator/ContractManagerEmulator.java
```diff
@@ -134,4 +141,20 @@ private Contract emulateContract(ContractEntry entry) throws Exception {
...
     return contract;
   }
+
+  /**
+   * Emulate the certificate holderId and version for the given Contract instance
+   *
+   * @param contract the contract in which the certificate will be emulated
+   */
+  private void emulateCertificateKey(Contract contract) {
+    if (emulatedCertificateKey == null) {
```
feeblefakie: I'm wondering what would happen if `set-certificate` is not called and getCertificate() is called.
Does it fail ?
If so, maybe it might be good to set the default certificate since real contracts always can execute `getCertificate`.

emulator/src/test/java/com/scalar/client/tool/emulator/Callee.java
```diff
@@ -25,10 +25,19 @@
...
 import javax.json.Json;
 import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
 
 public class Callee extends Contract {
   @Override
   public JsonObject invoke(Ledger ledger, JsonObject argument, Optional<JsonObject> properties) {
-    return Json.createObjectBuilder().add("callee_is_called", true).build();
+    JsonObjectBuilder result = Json.createObjectBuilder();
+    if (getCertificateKey() != null) {
```
feeblefakie: In a real contract, it never returns null, so it might be better to set the default certificate as I suggested above.

emulator/src/main/java/com/scalar/client/tool/emulator/ContractManagerEmulator.java
Torch3333: If `set-certificate` is not called, `getCertificate()` will return `null`. 
Indeed, you are right. Should I set a default value for the certificate such as ? 
- `holderId` = `default_holder_id`
- `version` = `1`

feeblefakie: Yes, let's do that way.

Torch3333: The changes are in commit https://github.com/scalar-labs/scalardl-tools/pull/24/commits/bbee3885170199aa4ce93d93929c6e1b9695e806

PR #35, opened 2023-04-25
Fix URLs in README

josh-wong:
This PR fixes the relative URLs that point to the tool folders in `docs/README.md`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,7 +2,7 @@
 
 This repository contains tools for use with Scalar DL.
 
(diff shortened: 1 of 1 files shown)


feeblefakie (commented):
Sorry, this repo is pretty obsolete. I think no one uses.
Emulator is not maintained any more so it should be deleted.
Explorer is something we can integrate into our command line tools.
But, anyway, I will archive the repo, so please don't use it for the doc site for now.

PR #37, opened 2026-05-08
Add Cosmos DB resumable scan

KodaiD:
## Description

Add a resumable parallel scanner for Cosmos DB tables used by ScalarDB.

This tool scans all physical partitions (FeedRanges) of Cosmos DB in parallel, checkpoints progress per page, and can resume from the last checkpoint after a failure.

## Related issues and/or PRs

N/A

## Changes made

  - Add `coordinator-state-deleter/resumable-scan` Gradle subproject with `ResumableScanner` interface and
   `CosmosResumableScanner` implementation
  - Implement per-FeedRange parallel scanning with page-level checkpoint persistence via
  `CosmosScanWorker` and `CheckpointManager`
  - Add unit tests and Cosmos DB integration tests
  - Add CI workflow for unit tests and Cosmos DB integration tests

## Checklist

- User-initiated pauses and resumes are out of scope.
- Checkpointing is performed at the page level. If a failure occurs during a page read, that page will be re-read from the beginning.
- Tests for partition-split cases. Will work on a separate PR.

## Release notes

Added Cosmos DB resumable scan.


coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScanner.java
```diff
@@ -0,0 +1,249 @@
...
+
+    if (feedRanges.isEmpty()) {
+      logger.info("No FeedRanges found for {}; nothing to scan", qualifiedTableName);
+      return new ScanResult(0);
+    }
+
+    // Shut down previous executor if scan() is called for multiple tables sequentially
+    shutdownExecutorIfExists(scanExecutor);
+    int threadCount = Math.min(feedRanges.size(), MAX_SCAN_THREADS);
+    scanExecutor = Executors.newFixedThreadPool(threadCount);
```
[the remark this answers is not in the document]
KodaiD: I've applied this in [cb9d257](https://github.com/scalar-labs/scalardl-tools/pull/37/commits/cb9d2578d0e6fa324391b8b4e0ac542537fe1d1f).

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/FeedRangeSerializer.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+class FeedRangeSerializer {
+
+  private FeedRangeSerializer() {}
+
+  /** Convert a FeedRange to a stable, filesystem-safe identifier. */
+  public static String toId(FeedRange feedRange) {
+    String json = feedRange.toString();
+    // Use a hash to produce a fixed-length filesystem-safe identifier
+    return Hashing.sha256().hashUnencodedChars(json).toString();
```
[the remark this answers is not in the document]
KodaiD: I've applied this in [cb9d257](https://github.com/scalar-labs/scalardl-tools/pull/37/commits/cb9d2578d0e6fa324391b8b4e0ac542537fe1d1f).

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScanner.java
```diff
@@ -0,0 +1,249 @@
...
+
+    logger.info("Cosmos DB physical partitions: {}", feedRanges.size());
+
+    if (feedRanges.isEmpty()) {
+      logger.info("No FeedRanges found for {}; nothing to scan", qualifiedTableName);
+      return new ScanResult(0);
+    }
+
+    // Shut down previous executor if scan() is called for multiple tables sequentially
+    shutdownExecutorIfExists(scanExecutor);
```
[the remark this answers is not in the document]
KodaiD: Extended the try block in [cb9d257](https://github.com/scalar-labs/scalardl-tools/pull/37/commits/cb9d2578d0e6fa324391b8b4e0ac542537fe1d1f).

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
```diff
@@ -0,0 +1,430 @@
...
+    }
+    return count;
+  }
+
+  /**
+   * Recreates the given container with dedicated throughput (container-level) that exceeds the
+   * per-partition threshold (~10,000 RU/s), ensuring Cosmos DB allocates multiple physical
+   * partitions.
+   */
+  private static void recreateContainerWithDedicatedThroughput() throws Exception {
```
jnmt, line 165: I understand this approach guarantees multiple physical partitions by provisioning a high RU throughput. I may not fully understand the billing model, but would this cost roughly ~$1 per CI run?

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CheckpointManager.java
```diff
@@ -0,0 +1,138 @@
...
+    this.checkpointDir = checkpointDir;
+  }
+
+  @SuppressFBWarnings(
+      value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
+      justification = "target.getFileName() is never null for our use cases (non-root paths)")
+  private static void writeAtomic(Path target, byte[] content) throws IOException {
+    Path parent = target.getParent();
+    if (parent != null) {
+      Files.createDirectories(parent);
```
jnmt: [minor performance improvement] `writeAtomic()` calls `Files.createDirectories(parent)` on every invocation, which means it runs on every page checkpoint. Since each scan worker writes to the same directory throughout its lifecycle, is it possible to create the directory once at worker startup (or lazily on first write) to avoid redundant system calls?

coordinator-state-deleter/resumable-scan/build.gradle
```diff
@@ -0,0 +1,74 @@
...
+}
+
+configurations {
+    integrationTestImplementation.extendsFrom testImplementation
+    integrationTestRuntimeOnly.extendsFrom testRuntimeOnly
+}
+
+dependencies {
+    api "com.azure:azure-cosmos:$azureCosmosVersion"
+    api "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
```
jnmt: `azure-cosmos` and `jackson-databind` are declared with `api` scope, but the public interface (`ResumableScanner`) doesn't expose types from these libraries. Would `implementation` be sufficient here?

coordinator-state-deleter/build.gradle
```diff
@@ -0,0 +1,93 @@
...
+
+    tasks.withType(Test).configureEach {
+        maxHeapSize "512m"
+        testLogging.showStandardStreams = true
+        useJUnitPlatform()
+    }
+
+    java {
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
```
jnmt, line 83: Recently, we have often used the toolchain to specify the Java version. Is my understanding correct that using the toolchain was avoided due to version compatibility with ErrorProne and SpotBugs? If we don't use the toolchain, it might be worth adding `options.release = 8` to verify that no newer APIs are accidentally used.

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
jnmt, line 1: If there's a missing test case, I think it would be verifying behavior when a physical partition split occurs mid-operation. While I assume there are no inserts or updates to the target table during execution of the deletion tool, I'd like to confirm that the scanner handles partition splits gracefully in case they occur for performance or other reasons. Specifically:
 
1. A scan in progress completes successfully even if the number of physical partitions increases during the scan.
2. After an interruption, checkpoints remain usable even if partitions have increased before resuming.                                                            

If this can be tested cost-effectively using the high provisioned RU approach, it might be worth adding as a test case. If it's difficult to automate or too costly/time-consuming to run on every CI execution, I'd like to verify this manually before proceeding with the implementation of other commands. What do you think?

jnmt (commented):
I think it's well-implemented overall. Thank you. Left a few comments. PTAL 🙇‍♂️

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/FeedRangeSerializer.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+class FeedRangeSerializer {
+
+  private FeedRangeSerializer() {}
+
+  /** Convert a FeedRange to a stable, filesystem-safe identifier. */
+  public static String toId(FeedRange feedRange) {
+    String json = feedRange.toString();
+    // Use a hash to produce a fixed-length filesystem-safe identifier
+    return Hashing.sha256().hashString(json, StandardCharsets.UTF_8).toString();
```
jnmt: [nitpick] Can we use `fromJson`?

```suggestion
    // Use a hash to produce a fixed-length filesystem-safe identifier
    return Hashing.sha256().hashString(fromJson(json), StandardCharsets.UTF_8).toString();
```

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CheckpointManager.java
```diff
@@ -0,0 +1,138 @@
...
+    Path feedRangesPath = checkpointDir.resolve(tableName).resolve(FEED_RANGES_FILE);
+    try {
+      writeAtomic(feedRangesPath, feedRangesJson.getBytes(StandardCharsets.UTF_8));
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to persist feed ranges for " + tableName, e);
+    }
+  }
+
+  /** Delete checkpoint files for the given table, keeping the checkpoint directory itself. */
+  public void clearTable(String tableName) {
```
jnmt: [nitpick] It doesn't actually clear the table, so I feel a bit odd. `clearCheckpointFiles(...)`, `clearCheckpointFor(String table)`, or something?

jnmt (commented):
Sorry, a few more minor comments. PTAL!

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
KodaiD: Good point. Cosmos DB bills using Request Units (RU) measured per second (RU/s). cf. https://azure.microsoft.com/en-us/pricing/details/cosmos-db/autoscale-provisioned/

This test creates a container with 12,000 RU/s for a user table and a container with 400 RU/s for a metadata table.

A container with 12,000 RU/s throughput will cost $0.016/m ($0.008 / 100RU/s / h), and a CI run takes about 2 minutes (It took 1m48s, previously). So, a CI run will cost about $0.032.

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CheckpointManager.java
KodaiD: Thank you for the suggestion! I've considered the approach, but I'd like to keep the current implementation based on the following points:

- The directory creation is an internal detail of CheckpointManager, so I'd like it to be triggered from within its own methods (`persistContinuationToken` and `persistFeedRanges`, or `writeAtomic`).
- `persistFeedRanges` is always called once per table before `persistContinuationToken`, so creating the directory there alone would be sufficient for now. However, having CheckpointManager rely on a specific call order feels fragile.
- `Files.createDirectories` is essentially a no-op when the directory already exists, so the performance overhead is minimal.

What do you think? I'd like to hear your opinion.

coordinator-state-deleter/resumable-scan/build.gradle
KodaiD: Thanks for pointing it out! Fixed in 79a1df8.

coordinator-state-deleter/build.gradle
KodaiD: Yes, it's due to the version incompatibility with ErrorProne (it requires Java 11+). I've added `options.release = 8` in 79a1df8. Thanks!

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CheckpointManager.java
KodaiD: Fixed in 79a1df8. Thanks!

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
KodaiD: Good point. Thank you for clarifying the situations that should be tested.

Currently, I'm working on more complex test cases, which include partition splits. They'll be added in a separate PR, even if it's possible to execute them cost-effectively in CI. Let me continue working on it and treat it out of scope in this PR.

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/FeedRangeSerializer.java
KodaiD: Maybe `toJson`? Fixed in 7789db9. Thanks!

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
jnmt: If I understand the following guide correctly, is the billing on an hourly basis?

https://learn.microsoft.com/en-us/azure/cosmos-db/understand-your-bill

> If you increase provisioned throughput at 9:30 AM from 400 RU/sec to 1,000 RU/sec and then lower provisioned throughput at 10:45 AM back to 400 RU/sec, you will be charged for two hours of 1,000 RU/sec.

coordinator-state-deleter/resumable-scan/build.gradle
```diff
@@ -0,0 +1,74 @@
...
+
+configurations {
+    integrationTestImplementation.extendsFrom testImplementation
+    integrationTestRuntimeOnly.extendsFrom testRuntimeOnly
+}
+
+dependencies {
+    implementation "com.azure:azure-cosmos:$azureCosmosVersion"
+    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
+    implementation "com.scalar-labs:scalardb:$scalarDbVersion"
```
brfrn169: For consistency:
```suggestion
    implementation "com.azure:azure-cosmos:${azureCosmosVersion}"
    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
    implementation "com.scalar-labs:scalardb:${scalarDbVersion}"
```

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
KodaiD: Sorry, it looks like I misread it. It seems we will be charged on an hourly basis. So, a CI run will cost:

12,000 RU/s: 120 × $0.008/100 RU/s/hour = $0.96/hour

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CheckpointManager.java
KodaiD: I've changed my mind. Since `clearCheckpointFor` exists, we can naturally provide `initCheckpointFor` as well. Fixed in 28992b0.

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
```diff
@@ -0,0 +1,430 @@
...
+   * #TEST_LOGICAL_PARTITION_COUNT} = 5, this gives 120 records per logical partition.
+   */
+  private static final int TEST_RECORD_COUNT = 600;
+
+  private static final int TEST_LOGICAL_PARTITION_COUNT = 5;
+
+  /**
+   * Throughput high enough to force at least 2 physical partitions (~10,000 RU/s per partition).
+   */
+  private static final int HIGH_THROUGHPUT_RU = 12_000;
```
KodaiD, line 74: Ref: https://learn.microsoft.com/en-us/azure/cosmos-db/partitioning#physical-partitions

> The amount of throughput provisioned (each individual physical partition can provide a throughput of up to 10,000 request units per second).

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CheckpointManager.java
jnmt: @KodaiD Thank you!

jnmt (approved):
LGTM, so far. Thank you! I will re-review when you add more integration test cases for the partition split (another PR is also fine).

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScanner.java
```diff
@@ -0,0 +1,254 @@
...
+          throw e;
+        }
+      }
+      if (firstFailure instanceof RuntimeException) {
+        throw (RuntimeException) firstFailure;
+      } else if (firstFailure instanceof Exception) {
+        throw (Exception) firstFailure;
+      } else if (firstFailure != null) {
+        throw new RuntimeException(firstFailure);
+      }
```
brfrn169: Nit: We should also handle `Error`:

```suggestion
      if (firstFailure instanceof Error) {
        throw (Error) firstFailure;
      } else if (firstFailure != null) {
        assert firstFailure instanceof Exception;
        throw (Exception) firstFailure;
      }
```

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScanner.java
```diff
@@ -0,0 +1,254 @@
...
+ * <p>Encapsulates all Cosmos-specific details: CosmosClient creation, namespace-to-database /
+ * table-to-container mapping, FeedRange discovery, per-partition parallel scanning, and
+ * raw-document-to-Result conversion.
+ */
+@NotThreadSafe
+public final class CosmosResumableScanner implements ResumableScanner {
+
+  private static final Logger logger = LoggerFactory.getLogger(CosmosResumableScanner.class);
+  private static final ObjectMapper mapper = new ObjectMapper();
+  private static final int MAX_SCAN_THREADS = 32;
```
brfrn169: Don't we need to make this configurable?

brfrn169 (commented):
Left several minor comments. PTAL!

coordinator-state-deleter/resumable-scan/build.gradle
KodaiD: Fixed in [a322a00](https://github.com/scalar-labs/scalardl-tools/pull/37/commits/a322a001cefb82467252e11dd444576db0dc1b65). Thanks!

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScanner.java
KodaiD: Fixed in [a322a00](https://github.com/scalar-labs/scalardl-tools/pull/37/commits/a322a001cefb82467252e11dd444576db0dc1b65). Thanks!

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScanner.java
KodaiD: It's better if it's configurable. Fixed in [a322a00](https://github.com/scalar-labs/scalardl-tools/pull/37/commits/a322a001cefb82467252e11dd444576db0dc1b65). Thanks!

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosScanWorker.java
```diff
@@ -0,0 +1,97 @@
...
+    this.resultInterpreter = resultInterpreter;
+    this.checkpointManager = checkpointManager;
+  }
+
+  @Override
+  public Long call() throws Exception {
+    long scanned = 0;
+
+    CosmosQueryRequestOptions options = new CosmosQueryRequestOptions();
+    options.setFeedRange(feedRange);
```
brfrn169, line 66: We can also set the max item count via `CosmosQueryRequestOptions.setMaxItemCount`. It might be better to make it configurable for performance as well.

brfrn169 (approved):
Left one more comment. Other than that, LGTM! Thank you!

coordinator-state-deleter/resumable-scan/src/main/java/com/scalar/dl/tools/scan/cosmos/CosmosScanWorker.java
KodaiD: It seems that CosmosQueryRequestOptions does not provide `setMaxItemCount()`. But we can specify the max item count in `CosmosPagedIterable<Record>.iterableByPage`. Let me think about this.

KodaiD: Make it configurable in be444ae. Thanks!

feeblefakie (approved):
LGTM! Thank you!

coordinator-state-deleter/resumable-scan/src/integration-test/java/com/scalar/dl/tools/scan/cosmos/CosmosResumableScannerIntegrationTest.java
feeblefakie: @jnmt @KodaiD Since it's in the scalardl-tools repo, the CI won't be run that often, so $1 per CI is probably fine.
But the CI will be triggered in each push, so let's be a bit careful.

PR #40, opened 2026-05-19
Add ledger finalization records logic

KodaiD:
## Description

This PR introduces a feature called ledger-finalize-records that scans all user tables and finalizes non-terminal (i.e., prepared or deleted) records by rolling them forward or back based on the coordinator state. This is the ledger-side counterpart required before coordinator state cleanup can proceed.

## Related issues and/or PRs

Depends on:

- #39  

## Changes made

  - Add `LedgerFinalizeOrchestrator` to discover tables, manage checkpointed state, and finalize records
  - Implement `RecordFinalizer` to roll forward or roll back non-terminal records
  - Implement `RecordStateChecker` to determine record finalization eligibility based on coordinator state and guarantee timestamp
  - Add `LedgerFinalizeState` and `LedgerFinalizeStateManager` for persisting and resuming finalization progress across restarts
  - Add `CompletionToken` and `StateManager` to the `common` module as shared utilities
  - Add unit tests
  - Add integration tests for the Cosmos DB implementation

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

- CLI part will be added in a separate PR.
- Integration tests with tables without a clustering key will be added in a separate PR.
- CI refactoring for cost efficiency will be addressed in a separate PR.
- User-facing error messages will be added in a separate PR.

## Release notes

Added ledger-finalize-records feature.

PR #41, opened 2026-06-05
Add coordinator state cleanup logic

KodaiD:
## Description

This PR introduces a feature called coordinator-state-cleanup that deletes records from the coordinator table that are safe to remove. It receives two completion tokens from ledger-finalize-records and auditor-finalize-records (will be added later).

## Related issues and/or PRs

Depends on:

- #40   

## Changes made

  - Add `CoordinatorCleanupOrchestrator` to compute the deletable-before timestamp, scan the coordinator table, and delete records
  - Implement `RecordDeleter` to delete coordinator records
  - Implement `CoordinatorCleanupState` and `CoordinatorCleanupStateManager` for resumable checkpoint persistence
  - Add unit tests
  - Add integration tests for the Cosmos DB implementation

## Checklist

- CLI part will be added in a separate PR.
- Integration tests with tables without a clustering key will be added in a separate PR.
- CI refactoring for cost efficiency will be addressed in a separate PR.
- User-facing error messages will be added in a separate PR.

## Release notes

Added coordinator-state-cleanup feature.


coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/RecordDeleter.java
```diff
@@ -0,0 +1,130 @@
...
+    this.deletedCount = new AtomicLong(0);
+  }
+
+  /**
+   * Submits a coordinator table record for deletion. Uses a semaphore for backpressure so that at
+   * most {@link #MAX_IN_FLIGHT} deletions are in flight at any time. The calling thread blocks on
+   * {@code permits.acquire()} when the limit is reached and resumes as soon as any single deletion
+   * completes.
+   */
+  public void submit(Result scanResult) throws Exception {
```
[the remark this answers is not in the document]
KodaiD: Fixed in [e86d440](https://github.com/scalar-labs/scalardl-tools/pull/41/commits/e86d440639adaab9beaa07574cd0b06927e8bc71).

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
```diff
@@ -0,0 +1,220 @@
...
+
+    try (ResumableScanner scanner = scannerFactory.create(scanCheckpointDir);
+        RecordDeleter deleter = recordDeleterFactory.create(storage, workerThreads)) {
+
+      ScanResult scanResult =
+          scanner.scan(
+              Coordinator.NAMESPACE,
+              Coordinator.TABLE,
+              result -> {
+                if (result.getBigInt(Attribute.CREATED_AT) < deletableBeforeMs) {
```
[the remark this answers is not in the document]
KodaiD: Fixed in [e86d440](https://github.com/scalar-labs/scalardl-tools/pull/41/commits/e86d440639adaab9beaa07574cd0b06927e8bc71).

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
```diff
@@ -0,0 +1,225 @@
...
+                if (!result.isNull(Attribute.CREATED_AT)
+                    && result.getBigInt(Attribute.CREATED_AT) < deletableBeforeMs) {
+                  try {
+                    deleter.submit(result);
+                  } catch (Exception e) {
+                    throw new RuntimeException(e);
+                  }
+                }
+              });
+      deleter.awaitCompletion();
```
jnmt: I think the checkpoint tracks *scan delivery*, not *delete durability*; `CosmosScanWorker` persists a page's continuation token right after submitting its records, while `RecordDeleter#submit` only enqueues the async delete. So a token can be persisted while its deletes are still in flight (or fail), and since `clearCheckpointFor()` only runs on success, a crash or mid-scan delete failure leaves the checkpoint ahead of the failed page, which is skipped on resume. So, a fresh full scan eventually is required to reclaim leftovers.

Do we need to have a page-based callback, e.g., `Consumer<List<Result>>` in addition to (or instead of) `Consumer<Result>`? Or should we run deletion synchronously?

jnmt (commented):
Thank you for the PR. We need to make an important decision on the efficient deletion. PTAL at my comment for details.

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
KodaiD: Separating the delete and finalize processes from the scan process might be an option. The scan process simply records the records to be cleaned up.

As you mentioned, deleting the state might also be appropriate from a RU (Rules Request) perspective.

Let's discuss this.

KodaiD: @jnmt and I discussed this and decided to go with the synchronous approach so far. Let's reconsider after evaluation if necessary.

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
```diff
@@ -0,0 +1,255 @@
...
+        Instant.ofEpochMilli(ledgerTimestamp),
+        Instant.ofEpochMilli(auditorTimestamp));
+
+    return deletableBeforeMs;
+  }
+
+  /** Scans the coordinator table and deletes deletable records. */
+  private void scanAndDelete(CoordinatorCleanupStateManager stateManager, long deletableBeforeMs)
+      throws Exception {
+    Path scanCheckpointDir = stateManager.getStateDir();
```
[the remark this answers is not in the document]
KodaiD: I'd like to revisit this once all three commands are complete.

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/RecordDeleter.java
```diff
@@ -0,0 +1,40 @@
...
+        result
+            .getPartitionKey()
+            .orElseThrow(() -> new IllegalArgumentException("Partition key not found in result"));
+
+    return Delete.newBuilder()
+        .namespace(coordinatorNamespace)
+        .table(Coordinator.TABLE)
+        .partitionKey(partitionKey)
+        .build();
+  }
```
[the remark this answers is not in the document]
KodaiD: Let's use the more straightforward API for now.

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
```diff
@@ -0,0 +1,255 @@
...
+          storage,
+          scannerFactory,
+          checkpointDir,
+          resolveCoordinatorNamespace(dbConfig),
+          ledgerTokenString,
+          auditorTokenString);
+    } catch (Exception e) {
+      storage.close();
+      throw e;
+    }
```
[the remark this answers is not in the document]
KodaiD: I'll handle this in a separate PR later.

KodaiD:
@jnmt @feeblefakie @brfrn169 
Updated this PR as well as the ledger-side-command PR. PTAL again!

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
```diff
@@ -0,0 +1,255 @@
...
+                  try {
+                    recordDeleter.execute(result);
+                  } catch (Exception e) {
+                    throw new RuntimeException(e);
+                  }
+                }
+              });
+
+      logger.info(
+          "Finished scanning the coordinator table. {} records were scanned.",
```
feeblefakie: Should it also log how many were deleted?

coordinator-state-deleter/coordinator-state-cleanup/src/integration-test/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestratorIntegrationTestBase.java
```diff
@@ -0,0 +1,272 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public abstract class CoordinatorCleanupOrchestratorIntegrationTestBase {
```
feeblefakie, line 36: Can we add a test that interrupts the orchestrator partway through a scan (e.g. throw from the delete on the Nth record) and then re-runs it, asserting that exactly the deletable records end up absent, the rest present, and the checkpoint finishes completed=true?
The unit test covers delete-failure propagation, but not the resume that follows it.

feeblefakie (commented):
Overall, looking good. Left some comments. PTAL!

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
KodaiD: @feeblefakie (cc @jnmt)

Currently, we are discussing this point. It's a little bit difficult to keep track of the correct deletion count over failures.

To keep track of the correct deletion count, we need to checkpoint the count every time a deletion succeeds, which introduces high overhead. Logging approximate counts can be one option, but it may confuse users. 

Do you have any idea?

KodaiD: I am considering the following options:

1) Log every time a deletion succeeds:
  - Pros: Users can get the exact deletion count by aggregating the logs, and it shows progress to the users.
  - Cons: Since the number of deletions is massive, the log size will be too big.
2) Checkpoint the deletion count:
  - Pros: The exact deletion count can be tracked even if the process is interrupted.
  - Cons: The latest counter value must be persisted every time a deletion succeeds, leading to high performance overhead.
3) Periodically log the deletion count for the current run like "500 records were deleted in this run":
  - Pros: It shows progress to the users while preventing the log size from bloating.
  - Cons: We need to decide the parameter for the logging frequency.

jnmt: @KodaiD Thanks for the discussion earlier. Right, if we want to do efficient checkpointing on a per-page or per-N-records basis, we won't be able to correctly count the records that were already deleted between the last checkpoint and the point of failure.

Logging or checkpointing per record isn't realistic either, so I think Option 3 is the way to go. I think it's sufficient to log successful deletions in batches of N records, and then display the total number of records deleted in this run at the end. That way, even if it fails partway through, we can still tell roughly how many records were deleted up to that point from the logs.

@feeblefakie What do you think?

coordinator-state-deleter/coordinator-state-cleanup/src/integration-test/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestratorIntegrationTestBase.java
KodaiD: Good catch! Thank you! Added in [8330712](https://github.com/scalar-labs/scalardl-tools/pull/41/commits/8330712283c90799df9e18a70d55b4322b6b9c65).
I'm planning to add more complex test cases after all the subcommands are ready.

coordinator-state-deleter/coordinator-state-cleanup/src/main/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestrator.java
feeblefakie: @KodaiD Thanks for the comments.
It's fine not to report the total number of deleted records, as long as it shows how many records are deleted in each run (command execution).
So, can we show the information like this? (the format is arbitrary)

```
Scanned records in total: xxx
Scanned records in this run: yyy
Deleted records in this run: zzz
```

KodaiD: @feeblefakie @jnmt  Thank you for the confirmation. Fixed in [fa1d356](https://github.com/scalar-labs/scalardl-tools/pull/41/commits/fa1d356f7e5be4d0b6107154da947965a5e794cb). PTAL!

`ScanResult.getTotalScanned()` actually returns a count in a run. It's a little bit confusing. I'll rename it in a subsequent refactoring PR.

jnmt (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

coordinator-state-deleter/coordinator-state-cleanup/src/test/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestratorTest.java
```diff
@@ -0,0 +1,405 @@
...
+            Coordinator.NAMESPACE,
+            ledgerToken,
+            auditorToken,
+            mockDeleter);
+
+    // Act & Assert
+    assertThatThrownBy(orchestrator::execute)
+        .isInstanceOf(RuntimeException.class)
+        .hasMessageContaining("DB unavailable");
+    verify(mockDeleter, never()).execute(record2);
```
brfrn169, line 244: This test verifies that the exception propagates and that the record after the failing one is not processed, but it does not assert that the checkpoint is left **resumable** — i.e. that the run did not get marked completed. Since resumability is the core contract of this tool, it's worth pinning down so a future regression (e.g. `markCompleted()` accidentally moving before the scan) is caught.

Suggested addition after the existing asserts:

```java
CoordinatorCleanupState state = new CoordinatorCleanupStateManager(tempDir).load();
assertThat(state).isNotNull();
assertThat(state.isCompleted()).isFalse();
```

coordinator-state-deleter/coordinator-state-cleanup/src/test/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestratorTest.java
```diff
@@ -0,0 +1,405 @@
...
+    CoordinatorCleanupOrchestrator orchestrator =
+        new CoordinatorCleanupOrchestrator(
+            storage, scannerFactory, tempDir, Coordinator.NAMESPACE, null, null);
+
+    // Act & Assert
+    assertThatThrownBy(orchestrator::execute)
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining(
+            "Both ledger and auditor completion tokens are required for the initial run");
+  }
```
brfrn169: This covers the both-tokens-null case, but the guard is `ledgerTokenString == null || auditorTokenString == null` — the exactly-one-null combinations are a realistic operator mistake (e.g. forgetting one of the two tokens) and are currently untested. Consider parameterizing this test to also cover `(ledgerToken, null)` and `(null, auditorToken)`, asserting the same IllegalArgumentException. Cheap, safe coverage of an actual error branch.

brfrn169 (approved):
Left a couple of minor comments for the unit tests. Other than that, LGTM! Thank you!

coordinator-state-deleter/coordinator-state-cleanup/src/test/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestratorTest.java
KodaiD: Good catch. Thanks! Fixed in [5b4e81f](https://github.com/scalar-labs/scalardl-tools/pull/41/commits/5b4e81f8366486ee90749c4cdabd387f2d1b4525).

coordinator-state-deleter/coordinator-state-cleanup/src/test/java/com/scalar/dl/tools/cleanup/CoordinatorCleanupOrchestratorTest.java
KodaiD: Fixed in [5b4e81f](https://github.com/scalar-labs/scalardl-tools/pull/41/commits/5b4e81f8366486ee90749c4cdabd387f2d1b4525). Thanks!

PR #59, opened 2026-07-21
Add Docker build configuration for ScalarDL Cleanup CLI

KodaiD:
## Description

This PR adds a Docker build configuration so the `scalardl-cleanup` CLI tools can be distributed and run as a container image on `ghcr.io/scalar-labs/scalardl-cleanup`.

## Related issues and/or PRs

N/A

## Changes made

- Add a `Dockerfile` that runs as a non-root `scalar` user (uid/gid 201) and unpacks the distribution tarball into `/scalar/scalardl-cleanup`.
- Add `docker-entrypoint.sh` that sizes the JVM heap from the container memory limit and execs the CLI launcher.
- Add Gradle `docker` and `copyFilesToDockerBuildContextDir` tasks, with a predictable `distTar` archive name and a `-PdockerVersion` override.

## Checklist

A release setting will be added in a separate PR.

## Release notes

N/A

### DIFF
--- a/scalardl-cleanup/cli/Dockerfile
+++ b/scalardl-cleanup/cli/Dockerfile
@@ -1,0 +1,39 @@
+FROM eclipse-temurin:21-jre-jammy
+
+RUN apt-get update && apt-get upgrade -y --no-install-recommends \
+ && apt-get autoremove -y \
+ && rm -rf /var/lib/apt/lists/* \
+ && groupadd -r --gid 201 scalar \
+ && useradd -r --uid 201 -g scalar scalar
+
+WORKDIR /scalar
+
+COPY scalardl-cleanup.tar docker-entrypoint.sh /tmp/build/
+RUN mkdir -p /scalar/scalardl-cleanup \
+ && tar -xf /tmp/build/scalardl-cleanup.tar -C /scalar/scalardl-cleanup --strip-components=1 \
+ && cp /tmp/build/docker-entrypoint.sh /scalar/scalardl-cleanup/ \
+ && chmod +x /scalar/scalardl-cleanup/docker-entrypoint.sh \
+ && rm -rf /tmp/build \
+ && chown -R 201:201 /scalar
+
+USER 201
+
+ENV JAVA_OPT_MAX_RAM_PERCENTAGE='60.0'
+
+ARG IMAGE_NAME
+ARG IMAGE_VERSION
+ARG IMAGE_RELEASE
+
+LABEL name="${IMAGE_NAME}" \
+      maintainer="Scalar, Inc." \
+      vendor="Scalar, Inc." \
+      version="${IMAGE_VERSION}" \
+      release="${IMAGE_RELEASE}" \
+      summary="ScalarDL Cleanup Tools Container Image" \
+      description="Command-line tools that let ScalarDL operators safely reclaim space by removing stale records left behind by completed transactions from ScalarDB's coordinator.state table (Ledger) and the Auditor's asset_lock/request_proof tables."
+
+WORKDIR /scalar/scalardl-cleanup
+
+ENTRYPOINT ["./docker-entrypoint.sh"]
+
+CMD ["--help"]
--- a/scalardl-cleanup/cli/Dockerfile
+++ b/scalardl-cleanup/cli/Dockerfile
@@ -30,7 +30,7 @@ LABEL name="${IMAGE_NAME}" \
       version="${IMAGE_VERSION}" \
       release="${IMAGE_RELEASE}" \
       summary="ScalarDL Cleanup Tools Container Image" \
-      description="Command-line tools that let ScalarDL operators safely reclaim space by removing stale records left behind by completed transactions from ScalarDB's coordinator.state table (Ledger) and the Auditor's asset_lock/request_proof tables."
+      description="Command-line tools that let ScalarDL operators safely reclaim space by removing stale records left behind by completed transactions from Ledger's coordinator table and the Auditor's request proof table."
 
 WORKDIR /scalar/scalardl-cleanup
 
--- a/scalardl-cleanup/cli/Dockerfile
+++ b/scalardl-cleanup/cli/Dockerfile
@@ -16,6 +16,9 @@ RUN apt-get update && apt-get upgrade -y --no-install-recommends \
 
 WORKDIR /scalar
 
+# Deliberately left owned by root:root (COPY's default — no --chown): uid 201 gets
+# read/execute but cannot write, so a compromised JVM process cannot rewrite the
+# entrypoint or the jars it is about to load.
 COPY --from=builder /out /scalar/scalardl-cleanup
 
 USER 201
--- a/scalardl-cleanup/cli/Dockerfile
+++ b/scalardl-cleanup/cli/Dockerfile
@@ -1,3 +1,11 @@
+FROM eclipse-temurin:21-jre-jammy AS builder
+
+COPY scalardl-cleanup.tar docker-entrypoint.sh /tmp/build/
+RUN mkdir -p /out \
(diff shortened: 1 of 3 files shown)


scalardl-cleanup/cli/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ARG IMAGE_VERSION
+ARG IMAGE_RELEASE
+
+LABEL name="${IMAGE_NAME}" \
+      maintainer="Scalar, Inc." \
+      vendor="Scalar, Inc." \
+      version="${IMAGE_VERSION}" \
+      release="${IMAGE_RELEASE}" \
+      summary="ScalarDL Cleanup Tools Container Image" \
+      description="Command-line tools that let ScalarDL operators safely reclaim space by removing stale records left behind by completed transactions from ScalarDB's coordinator.state table (Ledger) and the Auditor's asset_lock/request_proof tables."
```
jnmt: nitpick: The original sentence looks like removing asset lock entries. Also, maybe we don't have to say "ScalarDB's" for end-user friendliness and consistency (Ledger vs. Auditor).

```suggestion
      description="Command-line tools that let ScalarDL operators safely reclaim space by removing stale records left behind by completed transactions from Ledger's coordinator.state table and the Auditor's request_proof table."
```

jnmt (commented):
Overall looks good, thank you! Left one minor suggestion. PTAL.

scalardl-cleanup/cli/Dockerfile
KodaiD: Good catch. Fixed in [69d9eb3](https://github.com/scalar-labs/scalardl-tools/pull/59/commits/69d9eb30c6af5512f8a3d99bfdf0b8fff351928a) based on your suggestion. Thanks!

brfrn169 (approved):
LGTM! Thank you!

scalardl-cleanup/cli/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ && useradd -r --uid 201 -g scalar scalar
+
+WORKDIR /scalar
+
+COPY scalardl-cleanup.tar docker-entrypoint.sh /tmp/build/
+RUN mkdir -p /scalar/scalardl-cleanup \
+ && tar -xf /tmp/build/scalardl-cleanup.tar -C /scalar/scalardl-cleanup --strip-components=1 \
+ && cp /tmp/build/docker-entrypoint.sh /scalar/scalardl-cleanup/ \
+ && chmod +x /scalar/scalardl-cleanup/docker-entrypoint.sh \
+ && rm -rf /tmp/build \
```
feeblefakie: Docker layer semantics mean the `COPY` on line 11 permanently bakes `/tmp/build/scalardl-cleanup.tar` into that layer's blob — the `rm -rf /tmp/build` on line 16 only writes a whiteout in the next layer, it can't shrink the earlier one. Result: the app payload ships in the pulled image twice (once as the tar blob, once as the extracted tree), roughly doubling the pull size.

Standard fixes:
- Multi-stage: extract in a builder stage, `COPY --from=builder /out /scalar/scalardl-cleanup` in the final image — the tar never enters a final-image layer.
- BuildKit mount: `RUN --mount=type=bind,source=scalardl-cleanup.tar,target=/tmp/x.tar tar -xf /tmp/x.tar ...` — the tar stays a build-time mount.

Noting this is consistent with the sibling ledger image pattern, so it's a systemic tidy-up more than a defect specific to this PR — but this is the natural moment to fix it before the pattern replicates further.

scalardl-cleanup/cli/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ && chown -R 201:201 /scalar
```
feeblefakie: `chown -R 201:201 /scalar` makes every file — including `docker-entrypoint.sh` and `bin/scalardl-cleanup` — writable by the account the JVM runs as. If the JVM process is ever compromised (deserialization gadget, path-traversal write, etc.), the attacker can rewrite the entrypoint and backdoor subsequent invocations. Standard hardening: leave `/scalar` owned by `root:root` (files 0644, scripts 0755) so UID 201 keeps read/execute but not write. Only chown dirs the tool actually writes to.

feeblefakie (commented):
Left some comments. PTAL!

scalardl-cleanup/cli/Dockerfile
KodaiD: Good catch. Fixed with a multi-stage build in [c89b293](https://github.com/scalar-labs/scalardl-tools/pull/59/commits/c89b293f6cdf0d6e5853b882a13354c0c4528c90). Thanks!

scalardl-cleanup/cli/Dockerfile
KodaiD: Applied in c89b293. Dropped the chown, so `/scalar` stays `root:root` and UID 201 keeps read/execute only. No pre-created writable dir in the image either — the checkpoint always goes to a mounted volume, never to the container's own filesystem.

feeblefakie (approved):
LGTM! Thank you!

jnmt (approved):
LGTM, thank you!

scalardl-cleanup/cli/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+RUN apt-get update && apt-get upgrade -y --no-install-recommends \
+ && apt-get autoremove -y \
+ && rm -rf /var/lib/apt/lists/* \
+ && groupadd -r --gid 201 scalar \
+ && useradd -r --uid 201 -g scalar scalar
+
+WORKDIR /scalar
+
+COPY --from=builder /out /scalar/scalardl-cleanup
```
kota2and3kan, line 22:
```suggestion
# To avoid security issues like updating (writing) binary files by the application process if the JVM process is ever compromised, we intentionally set the `user:group` configuration to `root:root` here.
COPY --from=builder /out /scalar/scalardl-cleanup
```
I think it might be better to add a comment to explain the reason why we use `root:root` here.

(Also, it might be better to apply the same configuration to other tool container images in the future. Let me think about it later.)

scalardl-cleanup/cli/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ARG IMAGE_VERSION
+ARG IMAGE_RELEASE
+
+LABEL name="${IMAGE_NAME}" \
+      maintainer="Scalar, Inc." \
+      vendor="Scalar, Inc." \
+      version="${IMAGE_VERSION}" \
+      release="${IMAGE_RELEASE}" \
+      summary="ScalarDL Cleanup Tools Container Image" \
+      description="Command-line tools that let ScalarDL operators safely reclaim space by removing stale records left behind by completed transactions from Ledger's coordinator table and the Auditor's request proof table."
```
kota2and3kan, line 38: Just a note. For now, basically, we don't set those label configurations in the tools containers. We will consider adding those label configurations to container images of other Scalar products tools, for example, SQL CLI, Schema Loader, and so on.
(This comment does not prevent merging this PR since this is just a reminder for future developers.)

kota2and3kan (commented):
Thank you for the updates!
Overall looks good to me, but I left several minor comments.
Please take a look when you have time! 🙇
(As we discussed in the meeting today, you can skip my approval for now if the release is a priority.)

scalardl-cleanup/cli/Dockerfile
KodaiD: Added a comment in [54a32ae](https://github.com/scalar-labs/scalardl-tools/pull/59/commits/54a32ae375d8f2b68b532bb6daa22c90322e5e2f). Thanks!

KodaiD:
@kota2and3kan Thank you for your review!
I'll merge this PR. Please let me know if you notice anything when you get back.

PR #62, opened 2026-07-24
Enable TLS on the Auditor server in the E2E test

KodaiD:
## Description

This PR enables TLS on the Auditor server in the E2E test to exercise `finalize-auditor` against a TLS-enabled Auditor. The change will help with verification when minimizing the properties passed to `finalize-auditor` in a subsequent PR.

## Related issues and/or PRs

N/A

## Changes made

- Generate a disposable self-signed Auditor TLS cert/key at deploy time and publish it as the `auditor-tls` Secret.
- Enable `scalar.dl.auditor.server.tls.*` in `auditor.yaml` and mount the cert/key at `/etc/scalardl-tls`.
- Configure the clients to trust the cert and pin its SAN via `scalar.dl.client.auditor.tls.*` in `client.properties`.
- Update the Auditor gRPC health probes to speak TLS and verify against the SAN.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [ ] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

N/A

## Release notes

N/A

PR #63, opened 2026-07-24
Simplify finalize-auditor configuration to a single properties file

KodaiD:
## Description

This PR simplifies the configuration of the `finalize-auditor`. Because the command communicates with the Auditor server, it requires some client properties. Currently, the command requires a full client configuration, but it's too much. This PR identifies what's really needed and makes the required configuration minimal.

## Related issues and/or PRs

We should test this change with:

- #62 

## Changes made

- Unify `AuditorFinalizeOrchestrator.create()` to take a single `Properties` instead of separate Auditor and client properties.
- Add `AuditorFinalizeConfig` to resolve the Auditor server connection (host, ports, TLS) and base namespace from that single properties file.
- Build `AuditorClient` from a connection target only, dropping the need for a full ScalarDL client configuration.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
+++ b/scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
@@ -1,0 +1,69 @@
+package com.scalar.dl.tools.cleanup;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.scalar.dl.client.config.ClientConfig;
+import com.scalar.dl.ledger.config.ConfigUtils;
+import com.scalar.dl.ledger.config.GrpcClientConfig;
+import com.scalar.dl.ledger.config.TargetConfig;
+import com.scalar.dl.tools.common.AuditorInternalValues;
+import java.util.Properties;
+
+/** Auditor-tool-specific settings for {@code auditor-finalize-records}. */
+public final class AuditorFinalizeConfig {
+
+  // Mirror ClientConfig.DEFAULT_* since they are private.
+  @VisibleForTesting static final long DEFAULT_GRPC_DEADLINE_MS = 60000;
+  @VisibleForTesting static final String DEFAULT_AUDITOR_HOST = "localhost";
+  @VisibleForTesting static final int DEFAULT_AUDITOR_PORT = 40051;
+  @VisibleForTesting static final int DEFAULT_AUDITOR_PRIVILEGED_PORT = 40052;
+  @VisibleForTesting static final boolean DEFAULT_AUDITOR_TLS_ENABLED = false;
+
+  private final String baseNamespace;
+  private final TargetConfig auditorTargetConfig;
+
+  public AuditorFinalizeConfig(Properties props) {
+    this.baseNamespace =
+        props.getProperty(
+            AuditorInternalValues.AUDITOR_NAMESPACE_PROPERTY,
+            AuditorInternalValues.DEFAULT_BASE_NAMESPACE);
+    this.auditorTargetConfig = buildAuditorTargetConfig(props);
+  }
+
+  @VisibleForTesting
+  static TargetConfig buildAuditorTargetConfig(Properties props) {
+    // The CA root cert is taken from the inline PEM if set, otherwise read from the cert file path.
+    String caRootCertPem =
+        ConfigUtils.getString(props, ClientConfig.AUDITOR_TLS_CA_ROOT_CERT_PEM, null);
+    String caRootCert =
+        caRootCertPem != null
+            ? caRootCertPem
+            : ConfigUtils.getStringFromFilePath(
+                props, ClientConfig.AUDITOR_TLS_CA_ROOT_CERT_PATH, null);
+
+    return TargetConfig.newBuilder()
+        .host(ConfigUtils.getString(props, ClientConfig.AUDITOR_HOST, DEFAULT_AUDITOR_HOST))
+        .port(ConfigUtils.getInt(props, ClientConfig.AUDITOR_PORT, DEFAULT_AUDITOR_PORT))
+        .privilegedPort(
+            ConfigUtils.getInt(
+                props, ClientConfig.AUDITOR_PRIVILEGED_PORT, DEFAULT_AUDITOR_PRIVILEGED_PORT))
+        .tlsEnabled(
+            ConfigUtils.getBoolean(
+                props, ClientConfig.AUDITOR_TLS_ENABLED, DEFAULT_AUDITOR_TLS_ENABLED))
+        .tlsCaRootCert(caRootCert)
+        .tlsOverrideAuthority(
+            ConfigUtils.getString(props, ClientConfig.AUDITOR_TLS_OVERRIDE_AUTHORITY, null))
+        .grpcClientConfig(
+            GrpcClientConfig.newBuilder().deadlineDurationMillis(DEFAULT_GRPC_DEADLINE_MS).build())
+        .build();
+  }
+
+  /** Get the Auditor base namespace, default {@code auditor}. */
+  public String getBaseNamespace() {
+    return baseNamespace;
+  }
+
+  /** Get the config used to reach the Auditor server. */
+  public TargetConfig getAuditorTargetConfig() {
+    return auditorTargetConfig;
+  }
+}
--- a/scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
+++ b/scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
@@ -52,6 +52,8 @@ public final class AuditorFinalizeConfig {
         .tlsCaRootCert(caRootCert)
         .tlsOverrideAuthority(
             ConfigUtils.getString(props, ClientConfig.AUDITOR_TLS_OVERRIDE_AUTHORITY, null))
+        .authorizationCredential(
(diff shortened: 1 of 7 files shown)


scalardl-cleanup/ci/e2e/run-cleanup.sh
```diff
@@ -34,9 +34,9 @@ pf_wait 40051 40052
 # config secrets so the tool sees exactly the ScalarDB configuration each server uses.
 kubectl -n ledger-e2e  get secret ledger-config  -o jsonpath='{.data.ledger\.properties}'  | base64 -d > "$ledger_props"
 kubectl -n auditor-e2e get secret auditor-config -o jsonpath='{.data.auditor\.properties}' | base64 -d > "$auditor_props"
-# finalize-auditor additionally needs ScalarDL client properties to reach the Auditor server's
+# finalize-auditor additionally needs some ScalarDL client properties to reach the Auditor server's
 # privileged port. The leading newline guards against the Secret value lacking a trailing newline.
-{ echo; cat "$HERE/client.properties"; } >> "$auditor_props"
+{ echo; echo "scalar.dl.client.auditor.host=127.0.0.1"; } >> "$auditor_props"
```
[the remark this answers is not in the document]
KodaiD: I will handle this after one of the PRs is merged.

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
```diff
@@ -0,0 +1,69 @@
...
+            ConfigUtils.getInt(
+                props, ClientConfig.AUDITOR_PRIVILEGED_PORT, DEFAULT_AUDITOR_PRIVILEGED_PORT))
+        .tlsEnabled(
+            ConfigUtils.getBoolean(
+                props, ClientConfig.AUDITOR_TLS_ENABLED, DEFAULT_AUDITOR_TLS_ENABLED))
+        .tlsCaRootCert(caRootCert)
+        .tlsOverrideAuthority(
+            ConfigUtils.getString(props, ClientConfig.AUDITOR_TLS_OVERRIDE_AUTHORITY, null))
+        .grpcClientConfig(
+            GrpcClientConfig.newBuilder().deadlineDurationMillis(DEFAULT_GRPC_DEADLINE_MS).build())
```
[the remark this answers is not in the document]
KodaiD: Fixed in [ac1d845](https://github.com/scalar-labs/scalardl-tools/pull/63/commits/ac1d8453108b151e4249042388d48d50d6d91bbd).

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
```diff
@@ -0,0 +1,71 @@
...
+        .tlsEnabled(
+            ConfigUtils.getBoolean(
+                props, ClientConfig.AUDITOR_TLS_ENABLED, DEFAULT_AUDITOR_TLS_ENABLED))
+        .tlsCaRootCert(caRootCert)
+        .tlsOverrideAuthority(
+            ConfigUtils.getString(props, ClientConfig.AUDITOR_TLS_OVERRIDE_AUTHORITY, null))
+        .authorizationCredential(
+            ConfigUtils.getString(props, ClientConfig.AUDITOR_AUTHORIZATION_CREDENTIAL, null))
+        .grpcClientConfig(
+            GrpcClientConfig.newBuilder().deadlineDurationMillis(DEFAULT_GRPC_DEADLINE_MS).build())
```
feeblefakie, line 64: Setting `scalar.dl.client.grpc.deadline_duration_millis` in the properties file used to flow through `ClientConfig` and reach the auditor target, but here it's silently overridden by the hardcoded 60s (the test `grpcDeadlineSetInProps_shouldStillUseFixedDeadline` locks the behavior in without saying why). If keeping it fixed is intentional, could you add a short comment explaining why? Otherwise, honoring the property via `ConfigUtils.getLong(props, ClientConfig.GRPC_DEADLINE_DURATION_MILLIS, DEFAULT_GRPC_DEADLINE_MS)` would preserve the existing knob.

feeblefakie (commented):
Left some comments. PTAL!

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
KodaiD: It's intentional. This tool runs in the same cluster as the Auditor, so RecoverAssetLock is a low-latency RPC. I think 60s is sufficient. I prefer to avoid providing many configuration items for users, and I think this can be omitted. I added a comment in [e823feb](https://github.com/scalar-labs/scalardl-tools/pull/63/commits/e823feb896b0a6735876ab79cd23f75c3d1b9833). PTAL!

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
```diff
@@ -0,0 +1,73 @@
...
+        .tlsEnabled(
+            ConfigUtils.getBoolean(
+                props, ClientConfig.AUDITOR_TLS_ENABLED, DEFAULT_AUDITOR_TLS_ENABLED))
+        .tlsCaRootCert(caRootCert)
+        .tlsOverrideAuthority(
+            ConfigUtils.getString(props, ClientConfig.AUDITOR_TLS_OVERRIDE_AUTHORITY, null))
+        .authorizationCredential(
+            ConfigUtils.getString(props, ClientConfig.AUDITOR_AUTHORIZATION_CREDENTIAL, null))
+        // Fixed deadline, not configurable: this tool runs in the same cluster as the Auditor, so
+        // 60s is always ample for the synchronous RecoverAssetLock RPC.
```
jnmt: Since `GrpcClientConfig` also has `maxInboundMessageSize` and `maxInboundMetadataSize`, which are not configurable here either (they fall back to the gRPC defaults of 4 MiB / 8 KiB when unset), how about extending this comment to cover them as well? They limit the sizes of the response message and metadata received from the Auditor, and since `AssetLockRecoveryResponse` contains only a single result enum, the defaults are more than enough. For example:

```
// The gRPC client settings are fixed, not configurable. The deadline is 60s, which is always
// ample for the synchronous RecoverAssetLock RPC since this tool runs in the same cluster as
// the Auditor. The max sizes of the inbound message and metadata (i.e., the response and its
// headers/trailers from the Auditor) are left at the gRPC defaults (4 MiB and 8 KiB), which
// are more than enough since the response carries only a single result code.
```

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
jnmt: I'm fine with forcing the default for gRPC settings, but I've added a follow-up comment for the code comment for more clarification.

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
```diff
@@ -0,0 +1,73 @@
...
+import com.scalar.dl.ledger.config.ConfigUtils;
+import com.scalar.dl.ledger.config.GrpcClientConfig;
+import com.scalar.dl.ledger.config.TargetConfig;
+import com.scalar.dl.tools.common.AuditorInternalValues;
+import java.util.Properties;
+
+/** Auditor-tool-specific settings for {@code auditor-finalize-records}. */
+public final class AuditorFinalizeConfig {
+
+  // Mirror ClientConfig.DEFAULT_* since they are private.
```
jnmt, line 14: Not a blocker for this PR: instead of mirroring, it might be worth making the `DEFAULT_*` values in ScalarDL's `ClientConfig` public, like ScalarDB does in `DatabaseConfig`. That would let tools reference them directly and avoid the risk of the copies drifting apart if the defaults change. Since this is the only place that needs it for now, the current approach is fine; just a thought for the future.

jnmt (approved):
Left a few comment and a minor suggestion. Other than that, LGTM. Thank you!

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
KodaiD: I applied your suggestion in [371c63c](https://github.com/scalar-labs/scalardl-tools/pull/63/commits/371c63cad1b311290bfa755b2c3e2ccb952fcce7). Thanks!

scalardl-cleanup/auditor-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/AuditorFinalizeConfig.java
KodaiD: Applied this in [371c63c](https://github.com/scalar-labs/scalardl-tools/pull/63/commits/371c63cad1b311290bfa755b2c3e2ccb952fcce7).

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #66, opened 2026-07-30
Skip non-transactional tables when finalizing Ledger records

KodaiD:
## Description

This PR fixes finalize-ledger aborting with `DB-CORE-10030: The column tx_state does not exist`.

The command discovered every table in every namespace. It read a transaction state from each record, but most ScalarDL tables are created without transaction support, so the first non-transactional table it reached ended the run.

The bug stayed hidden because the tool ran on ScalarDB 4.0.0-SNAPSHOT, whose `getNamespaceNames()` returns nothing for a schema created by a 3.x schema loader, leaving the command with no table to scan at all.

## Related issues and/or PRs

N/A

## Changes made

- Restrict table discovery to Consensus Commit tables, and log every table skipped.
- Align the ScalarDB and ScalarDL dependencies with the 3.x line (3.19.0-SNAPSHOT / 3.14.0-SNAPSHOT), so discovery matches the schema the servers use, and the E2E test exercises the scan.
- Add unit tests for a non-transactional table and for a listed table without metadata.

## Checklist

N/A

## Release notes

N/A


scalardl-cleanup/ledger-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/LedgerFinalizeOrchestrator.java
```diff
@@ -167,14 +170,33 @@ private List<String> discoverTables() throws Exception {
     for (String namespace : admin.getNamespaceNames()) {
       for (String table : admin.getNamespaceTableNames(namespace)) {
         String qualifiedTable = ScalarDbUtils.getFullTableName(namespace, table);
-        if (!qualifiedTable.equals(coordinatorTable)) {
-          tables.add(qualifiedTable);
+        if (qualifiedTable.equals(coordinatorTable)) {
+          continue;
         }
+        if (!isTransactionTable(namespace, table)) {
```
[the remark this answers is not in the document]
KodaiD: Good catch, but this is not reachable for users: the tool has not been released yet.

scalardl-cleanup/build.gradle
```diff
@@ -9,16 +9,15 @@ subprojects {
...
-    // Overridable at release time with -PprojectVersion=<version>.
-    project.version = project.findProperty('projectVersion') ?: '4.0.0-SNAPSHOT'
+    project.version = '1.0.0-SNAPSHOT'
 
     ext {
         guavaVersion = '33.6.0-jre'
         slf4jVersion = '2.0.17'
         log4jVersion = '2.25.4'
         jacksonVersion = '2.21.3'
-        scalarDbVersion = '4.0.0-SNAPSHOT'
```
jnmt, line 20: With this change, a genuinely empty result should no longer be possible. That said, an operational mistake or something could still leave us with zero tables while the command succeeds and emits a completion token, which is dangerous: the token drives the coordinator state deletion, so lazy recovery would no longer be possible. Could we add a check for this case, just to be safe?

scalardl-cleanup/ledger-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/LedgerFinalizeOrchestrator.java
```diff
@@ -167,14 +170,33 @@ private List<String> discoverTables() throws Exception {
...
+  /**
+   * Returns whether the table is managed by the Consensus Commit transaction manager.
+   *
+   * @throws IllegalStateException if the table has no metadata even though it was just listed.
+   */
+  private boolean isTransactionTable(String namespace, String table) throws ExecutionException {
+    TableMetadata metadata = admin.getTableMetadata(namespace, table);
+    if (metadata == null) {
+      throw new IllegalStateException(
+          "Table metadata not found for " + ScalarDbUtils.getFullTableName(namespace, table));
```
jnmt, line 204: Is there any possibility for users to see this message? If so, assigning an error code is better.

jnmt (commented):
Overall looks good, thank you! Left a few comments. PTAL!

scalardl-cleanup/ledger-finalize-records/src/main/java/com/scalar/dl/tools/cleanup/LedgerFinalizeOrchestrator.java
KodaiD: Let's keep it as is, since this branch will not be taken except in special cases—such as when observing an intermediate DDL state.

scalardl-cleanup/build.gradle
KodaiD: Good point. Thanks!

Added validation to ensure there are at least two target tables in [91d59c1](https://github.com/scalar-labs/scalardl-tools/pull/66/commits/91d59c16403f1ef3ef75aa1da2a03465f3ca7d3b).

jnmt (approved):
LGTM, thank you!

brfrn169 (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #74, opened 2026-08-27
Extract E2E data population into script

KodaiD:
## Description

This PR moves the E2E data population from inline `run:` blocks in `e2e.yaml` into a script.

The scenario harness that follows has to set up the same data for each scenario run, and inline bash in a workflow step cannot be called from anywhere — only that step runs it once. This is a pure refactoring: no assertions are added and no behavior changes.

## Related issues and/or PRs

- #73 

## Changes made

- Add `ci/e2e/populate.sh` with the population steps extracted from `e2e.yaml`.
- Reduce the three workflow steps to one-line calls into it.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [ ] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

N/A

## Release notes

N/A

PR #75, opened 2026-08-28
Restructure the E2E test around scenario scripts

KodaiD:
## Description

This PR moves the E2E test's cleanup run and its post-cleanup verification into a scenario script, establishing the convention that the scenarios that follow will use one self-contained script per scenario, run as one workflow step. Apart from the verification fix below, no behavior changes.

## Related issues and/or PRs

- #74 

## Changes made

- Add `ci/e2e/scenarios/happy-path.sh` with the cleanup run and the post-cleanup verification, and reduce the two workflow steps to a single call into it.
- Fail the verification when the populated-assets metadata is missing, unreadable, or lists the wrong number of IDs.
- Add `pf_start_all()` utility to `port-forward.sh`.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/scalardl-cleanup/ci/e2e/run-cleanup.sh
+++ b/scalardl-cleanup/ci/e2e/run-cleanup.sh
@@ -1,84 +1,60 @@
 #!/usr/bin/env bash
 #
-# Run the three ScalarDL Cleanup commands against the deployed E2E cluster the way an operator does:
+# The three operator-facing ScalarDL Cleanup commands, each run as a Kubernetes Job from
+# scalardl-cleanup/manifests. Source it:
 #
-#   Step 1 (Ledger AD)  credentials Secret -> checkpoint PVC -> ConfigMap -> finalize-ledger Job
-#   Step 2 (Auditor AD) credentials Secret -> checkpoint PVC -> ConfigMap (+ TLS optional settings)
-#                       -> finalize-auditor Job
-#   Step 3 (Ledger AD)  cleanup-coordinator Job, with both completion tokens
+#   source ci/e2e/run-cleanup.sh
 #
 # Expects in the environment:
 #   CLEANUP_VERSION  tag of the scalardl-cleanup image
 #   COSMOSDB_SHELL   path to the Azure Cosmos DB Shell binary, used to count rows in Cosmos
 #   RUNNER_TEMP      scratch directory for the rendered ConfigMaps
-#   RECORD_COUNT     number of records populated per category
 # and a kubectl context with the ledger-e2e / auditor-e2e namespaces deployed.
 
-set -euo pipefail
+E2E_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$E2E_DIR/cleanup-jobs.sh"
 
-HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-source "$HERE/cleanup-jobs.sh"
+# Prepare what every step needs. Call once, before the first step.
+init_cleanup_commands() {
+  require_vars CLEANUP_VERSION COSMOSDB_SHELL RUNNER_TEMP
 
-require_vars CLEANUP_VERSION COSMOSDB_SHELL RUNNER_TEMP RECORD_COUNT
+  # The Job manifests take the image tag as ${CLEANUP_VERSION}; the namespace is not in them and is
+  # passed to kubectl instead. LEDGER_NS / AUDITOR_NS come from common.sh, via cleanup-jobs.sh.
+  export CLEANUP_VERSION
 
-# The Job manifests take the image tag as ${CLEANUP_VERSION}; the namespace is not in them and is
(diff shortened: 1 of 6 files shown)


scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
```diff
@@ -0,0 +1,71 @@
...
+source "$E2E/port-forward.sh"
+
+# client.properties resolves the Auditor CA cert relative to scalardl-cleanup/.
+cd "$E2E/../.."
+
+props="$E2E/client.properties"
+
+require_vars CLIENT RUNNER_TEMP RECORD_COUNT
+
+"$E2E/run-cleanup.sh"
```
jnmt: Could you say a bit more about how you intend the split between `run-cleanup.sh` and this script? Now that the workflow no longer calls `run-cleanup.sh` directly, this is its only caller, and the assertions sit on both sides: `run-cleanup.sh` keeps the `coordinator.state` count checks (lines 79-82), which are tied to `RECORD_COUNT` and to the exact data `populate.sh` creates, while the application-layer checks live here.

If the idea is for later scenarios to reuse `run-cleanup.sh`, would it make sense to move those `RECORD_COUNT`-dependent assertions into `happy-path.sh` and leave `run-cleanup.sh` as a plain procedure driver?

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
```diff
@@ -0,0 +1,71 @@
...
+# Scenario: the operator's cleanup procedure against a cluster holding both committed data and
+# stranded Auditor locks.
+#
+# It runs the three subcommands the way an operator deploys them -- as Kubernetes Jobs from
+# scalardl-cleanup/manifests, in the documented apply order, so the deployment procedure is under
+# test too:
+#   finalize-ledger      -> emits the Ledger completion token
+#   finalize-auditor     -> recovers the stranded asset locks via RecoverAssetLock and cleans up
+#                           request_proof; emits the Auditor completion token
+#   cleanup-coordinator  -> deletes the settled coordinator.state records
```
jnmt: Nit: This rundown of the three subcommands partially duplicates the header of `run-cleanup.sh`. Might be better to refactor it.

jnmt (commented):
Overall looks good, thank you! Left a few comments. PTAL!

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
KodaiD: You're right. Fixed in [d16cbcc](https://github.com/scalar-labs/scalardl-tools/pull/75/commits/d16cbcc7a677e75a9b42ad1381806b84fb7b6d5b). Thanks!

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
KodaiD: Thank you for your review. It was indeed half-baked. I've updated the code in [d16cbcc](https://github.com/scalar-labs/scalardl-tools/pull/75/commits/d16cbcc7a677e75a9b42ad1381806b84fb7b6d5b). PTAL again!

scalardl-cleanup/ci/e2e/run-cleanup.sh
```diff
@@ -1,84 +1,60 @@
...
+  # The Job manifests take the image tag as ${CLEANUP_VERSION}; the namespace is not in them and is
+  # passed to kubectl instead. LEDGER_NS / AUDITOR_NS come from common.sh, via cleanup-jobs.sh.
+  export CLEANUP_VERSION
+
+  init_manifests_workdir
+  load_ad_credentials
+}
+
+# Run finalize-ledger (Ledger AD). Echoes the completion token.
+run_finalize_ledger() {
```
jnmt, line 33: Since the token comes back on stdout, every call site is a command substitution, and bash doesn't honour `set -e` for non-final commands inside `var=$(func)` — a failing `apply_job` or `setup_ledger_ad` is silently skipped here, and we only notice when `wait_for_job` times out 600s later. Is my understanding correct?

jnmt (commented):
Thanks for the fix! Could you do one more pass over the diff for docs/comments the restructuring left behind? E.g.,

- The "A scenario is a self-contained script..." paragraph was dropped from happy-path.sh, so the convention this PR is meant to establish is no longer written down anywhere.
- The comment at e2e.yaml:74 still says COSMOSDB_SHELL is used by run-cleanup.sh, though count_cosmos_items moved to happy-path.sh.

scalardl-cleanup/ci/e2e/run-cleanup.sh
KodaiD: Oh, it's tricky... Really good point. Fixed in [2d7d12b](https://github.com/scalar-labs/scalardl-tools/pull/75/commits/2d7d12bd4179bf0ab33c5c66e8eb21cd2c5a8230). Thanks!

jnmt (approved):
LGTM, thank you!

feeblefakie (approved):
LGTM! Thank you!

PR #80, opened 2026-09-02
Add ScalarDL Cleanup documentation

KodaiD:
## Description

This PR adds the documentation for the ScalarDL Cleanup tool. The existing deployment guide under `manifests/` covered only the deploy steps, so this PR takes it as the base and completes it with an overview of the tool and the limitations that a deployment must satisfy.

## Related issues and/or PRs

N/A

## Changes made

- Add `RECOVER_ASSET_LOCK_RPC_UNSUPPORTED`, which names the missing RPC and tells the operator to upgrade the Auditor.
- Raise it from `LockFinalizer` when the exception chain carries gRPC's `UNIMPLEMENTED`, and let every other status propagate unchanged.
- Move `manifests/README.md` to the repository root and update it to be a complete document.

## Checklist

N/A

## Release notes

N/A


scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+  end
+  L1 -->|"Ledger token"| L3
+  A2 -->|"Auditor token"| L3
+```
+
+## Limitations
+
+- The tool supports **Azure Cosmos DB for NoSQL** only. The tool rejects other configurations including `multi-storage`.
+- Ledger must not share the Coordinator table that its ScalarDB manages with any other ScalarDB
+  instance.
```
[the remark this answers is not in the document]
KodaiD: It's identical to the one for `tx_state_management.enabled`.

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+Run the steps in order, from the `manifests/` directory (the file paths below are relative to it).
+**Steps 1 and 2 are independent** (either operator can go first, or in parallel); **step 3 needs
+both tokens.** Each step exports the variables it needs at the top; `CLEANUP_VERSION` is the image
+tag (`ghcr.io/scalar-labs/scalardl-cleanup:<CLEANUP_VERSION>`).
+
+### Prerequisites
+
+- A Kubernetes worker node has room for the **2 vCPU and 4 GiB of memory** that each Job requests
+  (`requests` equal `limits`).
+- `kubectl` and `envsubst` (from `gettext`) are installed on your machine.
```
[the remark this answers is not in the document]
KodaiD: We can omit this since AKS satisfies the condition.

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
+# ScalarDL Cleanup
+
+## What is ScalarDL Cleanup?
+
+ScalarDL Cleanup removes the residual transaction state that a ScalarDL deployment accumulated while
+it was running a version of ScalarDL that does not support purge, which is a version earlier than
+ScalarDL 3.14.0. The residual transaction state consists of the following:
```
josh-wong: [nit] Revising sentences in git is a bit tough when a paragraph is split across multiple lines. For example, if the text on one line becomes longer than the rest, it would look a little odd since the lines above/below are no longer roughly the same length.

I think writing it naturally as a paragraph is better (as shown below) since changes often apply to the paragraph as a whole rather than to a segmented paragraph.

```suggestion
ScalarDL Cleanup removes the residual transaction state that a ScalarDL deployment accumulated while it was running a version of ScalarDL that does not support purge, which is a version earlier than ScalarDL 3.14.0. The residual transaction state consists of the following:
```

However, since this is internal documentation, I'll leave the decision to adjust the formatting up to you.

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
+# ScalarDL Cleanup
+
+## What is ScalarDL Cleanup?
+
+ScalarDL Cleanup removes the residual transaction state that a ScalarDL deployment accumulated while
```
feeblefakie:
```suggestion
ScalarDL Cleanup is a command-line tool that removes the residual transaction state that a ScalarDL deployment accumulated while
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+- **Coordinator state records:** The transaction state records that ScalarDB manages in the
+  Coordinator table on the Ledger side.
+- **Request proofs:** The records that Auditor stores for each client request to detect Byzantine
+  faults.
+
+Because ScalarDL does not purge this state by default, it accumulates over time and consumes
+storage. ScalarDL 3.14.0 and later can purge it, but none of the purge options remove the state that
+an earlier version wrote, because that state lacks the information that purge requires. Use this
+tool to remove that state, once, after you upgrade an existing deployment and before you enable
```
josh-wong: I'm not sure what `once` means in this part. Does it mean the following?:

```suggestion
tool once to remove that state after you upgrade an existing deployment and before you enable
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+| [`scalar.dl.client.auditor.tls.enabled`](https://scalardl.scalar-labs.com/docs/latest/configurations/#auditortlsenabled) | Optional. |
+| [`scalar.dl.client.auditor.tls.ca_root_cert_path`](https://scalardl.scalar-labs.com/docs/latest/configurations/#auditortlsca_root_cert_path) | Optional. |
+| [`scalar.dl.client.auditor.tls.ca_root_cert_pem`](https://scalardl.scalar-labs.com/docs/latest/configurations/#auditortlsca_root_cert_pem) | Optional. An inline PEM, used in place of `ca_root_cert_path`. |
+| [`scalar.dl.client.auditor.tls.override_authority`](https://scalardl.scalar-labs.com/docs/latest/configurations/#auditortlsoverride_authority) | Optional. |
+| [`scalar.dl.client.auditor.authorization.credential`](https://scalardl.scalar-labs.com/docs/latest/configurations/#auditorauthorizationcredential) | Optional. |
+| `scalar.dl.tools.scan.cosmos.max_threads` | Optional. The scan parallelism. Defaults to `32`. |
+| `scalar.dl.tools.scan.cosmos.page_size` | Optional. The records fetched per query page. Defaults to `100`. |
+
+## Getting started
```
brfrn169: `Getting started` sounds like an introductory section that a reader goes through before the real usage, but this section is the operational procedure itself, and it is the main body of this document. The tool is run once against a production deployment, and the only sections that follow are failure recovery and teardown, so there is nothing that comes "after getting started".

How about naming it after what it does, for example `Running the cleanup`?


```suggestion
## Running the cleanup
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+Run the steps in order, from the `manifests/` directory (the file paths below are relative to it).
+**Steps 1 and 2 are independent** (either operator can go first, or in parallel); **step 3 needs
+both tokens.** Each step exports the variables it needs at the top; `CLEANUP_VERSION` is the image
+tag (`ghcr.io/scalar-labs/scalardl-cleanup:<CLEANUP_VERSION>`).
+
+### Prerequisites
```
brfrn169: Minor, and feel free to ignore: these are preconditions for the whole procedure rather than a part of it, and an operator checks them at the same time as `Limitations` — while deciding whether their deployment can run this tool at all. Would it be worth promoting this to a top-level section right after `Limitations`?

brfrn169 (approved):
Left a couple of minor comments. Other than that, LGTM! Thank you!

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Limitations
+
+- The tool supports **Azure Cosmos DB for NoSQL** only. The tool rejects other configurations including `multi-storage`.
+- Ledger must not share the Coordinator table that its ScalarDB manages with any other ScalarDB
+  instance.
+- [`tx_state_management.enabled`](https://scalardl.scalar-labs.com/docs/latest/configurations/#tx_state_managementenabled)
+  has never been enabled.
+- [`coordinator.group_commit.enabled`](https://scalardb.scalar-labs.com/docs/latest/configurations/#coordinatorgroup_commitenabled)
+  has never been enabled.
```
josh-wong:
```suggestion
- [`tx_state_management.enabled`](https://scalardl.scalar-labs.com/docs/latest/configurations/#tx_state_managementenabled) must have never been enabled.
- [`coordinator.group_commit.enabled`](https://scalardb.scalar-labs.com/docs/latest/configurations/#coordinatorgroup_commitenabled) must have never been enabled.
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Manifest files
+
+The manifests are in `manifests/`, split into one directory per administrative domain — each
```
josh-wong: [nit] Em dashes shouldn't be surrounded by spaces.

```suggestion
The manifests are in `manifests/`, split into one directory per administrative domain—each
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+## What you can configure
+
+Configuration falls into two groups: **deploy-time variables** you set in your shell, and
+the **settings each command reads** from its `.properties` file (supplied via the ConfigMap or the Secret).
+
+### 1. Deploy-time variables
+
+| Variable | Used in | Description |
+|----------------------------------|------------------------------------------|----------------------------------------------------------------------|
+| `CLEANUP_VERSION`                | the Job manifests                        | Image tag, `ghcr.io/scalar-labs/scalardl-cleanup:<CLEANUP_VERSION>` |
```
josh-wong:
```suggestion
| `CLEANUP_VERSION`                | The Job manifests                        | Image tag, `ghcr.io/scalar-labs/scalardl-cleanup:<CLEANUP_VERSION>` |
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+| `LEDGER_TOKEN` / `AUDITOR_TOKEN` | `ledger-ad/cleanup-coordinator.yaml` | The two completion tokens                                            |
+
+### 2. Settings the commands accept
+
+Each command reads these properties from the `.properties` file that its AD's ConfigMap supplies.
+`scalar.db.storage` is preset to `cosmos`. Each property name links to its entry in the ScalarDB or
+ScalarDL configuration reference, except the `scalar.dl.tools.*` properties, which belong to this
+tool.
+
+**Ledger AD** — `ledger-ad/configmap.yaml`, read by `finalize-ledger` and `cleanup-coordinator`:
```
josh-wong: [nit] This should be an en dash instead of an em dash since `**Ledger AD**` is kind of the sentence heading.

```suggestion
**Ledger AD** – `ledger-ad/configmap.yaml`, read by `finalize-ledger` and `cleanup-coordinator`:
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+| Property | Description |
+|---|---|
+| [`scalar.db.contact_points`](https://scalardb.scalar-labs.com/docs/latest/configurations/#storage-related-configurations) | Required. The Cosmos DB account URI. |
+| [`scalar.db.password`](https://scalardb.scalar-labs.com/docs/latest/configurations/#storage-related-configurations) | Required. The Cosmos DB key, supplied from the AD's Secret as `${env:SCALAR_DB_PASSWORD}`. |
+| [`scalar.db.consensus_commit.coordinator.namespace`](https://scalardb.scalar-labs.com/docs/latest/configurations/#coordinatornamespace) | Optional. Set it only if Ledger overrides the default `coordinator` namespace. |
+| `scalar.dl.tools.scan.cosmos.max_threads` | Optional. The scan parallelism. Defaults to `32`. |
+| `scalar.dl.tools.scan.cosmos.page_size` | Optional. The records fetched per query page. Defaults to `100`. |
+
+**Auditor AD** — `auditor-ad/configmap.yaml`, read by `finalize-auditor`:
```
josh-wong: [nit] Similar to [my other comment](https://github.com/scalar-labs/scalardl-tools/pull/80/changes#r3920449503).

```suggestion
**Auditor AD** – `auditor-ad/configmap.yaml`, read by `finalize-auditor`:
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+both tokens.** Each step exports the variables it needs at the top; `CLEANUP_VERSION` is the image
+tag (`ghcr.io/scalar-labs/scalardl-cleanup:<CLEANUP_VERSION>`).
+
+### Prerequisites
+
+- A Kubernetes worker node has room for the **2 vCPU and 4 GiB of memory** that each Job requests
+  (`requests` equal `limits`).
+- `kubectl` and `envsubst` (from `gettext`) are installed on your machine.
+
+### Step 1 — Ledger AD: `finalize-ledger`
```
josh-wong, line 126:
```suggestion
### Step 1 – Ledger AD: `finalize-ledger`
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+Because ScalarDL does not purge this state by default, it accumulates over time and consumes
+storage. ScalarDL 3.14.0 and later can purge it, but none of the purge options remove the state that
+an earlier version wrote, because that state lacks the information that purge requires. Use this
+tool to remove that state, once, after you upgrade an existing deployment and before you enable
+purge on it. For details about the residual transaction state and the purge options, see
+[Purge the Residual Transaction State](https://scalardl.scalar-labs.com/docs/latest/purge-residual-transaction-state/).
+
+The tool is packaged as the container image `ghcr.io/scalar-labs/scalardl-cleanup` and runs as
+one-shot Kubernetes `Job`s in the cluster that already runs Ledger and Auditor. You can run it while
```
josh-wong: I don't think we need to put "Jobs" in backticks here since its the name of the resource.

```suggestion
one-shot Kubernetes Jobs in the cluster that already runs Ledger and Auditor. You can run it while
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+Wait for the Job to complete, then read the completion token from the log of its succeeded Pod. The
+command prints its result as one JSON line:
+
+```json
+{"status_code":"OK","output":{"completion_token":"<base64url-encoded token>"}}
+```
+
+Keep the token in `completion_token` — it is the **Ledger token**.
```
josh-wong:
```suggestion
Keep the token in `completion_token`—it is the **Ledger token**.
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+If the command fails, it prints an error message in place of that result, and the Job retries it up
+to three times (`backoffLimit: 3` in the manifest), starting a new Pod each time. If none of those
+attempts succeeds, the Job fails. Deploying the Job again without deleting the checkpoint volume
+resumes the processing from the same checkpoint data. To start the processing over from the beginning
+instead, see [Re-running after a failure](#re-running-after-a-failure).
+
+### Step 2 — Auditor AD: `finalize-auditor`
```
josh-wong:
```suggestion
### Step 2 – Auditor AD: `finalize-auditor`
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+**c. Config.** Set `scalar.db.contact_points` and `scalar.dl.client.auditor.host` in
+`auditor-ad/configmap.yaml` (for TLS, also uncomment the TLS properties listed in the optional note
+below), then apply it:
+
+```bash
+kubectl -n "${AUDITOR_NS:?}" apply -f auditor-ad/configmap.yaml
+```
+
+**Optional — TLS to Auditor.** If TLS is enabled on Auditor, do this *before* step (d).
```
josh-wong:
```suggestion
**Optional – TLS to Auditor.** If TLS is enabled on Auditor, do this *before* step (d).
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+In `auditor-ad/configmap.yaml` (step c), uncomment:
+
+- `scalar.dl.client.auditor.tls.enabled=true`
+- `scalar.dl.client.auditor.tls.ca_root_cert_path=/cert/scalardl-cleanup/tls.crt`
+- `scalar.dl.client.auditor.tls.override_authority=<auditor-cert-cn-or-san>` — set the value; only if it differs from the host you connect to
```
josh-wong: Since this is a complete sentence, let's capitalize the first letter and end the sentence with punctuation.

```suggestion
- `scalar.dl.client.auditor.tls.override_authority=<auditor-cert-cn-or-san>` – Set the value only if it differs from the host you connect to.
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+Then create the cert Secret holding the CA root cert that signed the Auditor server cert.
+
+```bash
+kubectl -n "${AUDITOR_NS:?}" create secret generic scalardl-cleanup-cert --from-file=tls.crt=<ca.pem>
+```
+
+**Optional — Auditor authorization credential.** If Auditor requires an authorization credential, do this *before* step (d).
```
josh-wong:
```suggestion
**Optional – Auditor authorization credential.** If Auditor requires an authorization credential, do this *before* step (d).
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+```bash
+: "${CLEANUP_VERSION:?}" && envsubst < auditor-ad/finalize-auditor.yaml | kubectl -n "${AUDITOR_NS:?}" apply -f -
+```
+
+Wait for the Job to complete, then read the completion token from the log of its succeeded Pod. The
+command prints its result as one JSON line, in the same shape as the one that `finalize-ledger`
+prints in step 1.
+
+Hand the token in `completion_token` — the **Auditor token** — to the Ledger operator.
```
josh-wong:
```suggestion
Hand the token in `completion_token`—the **Auditor token**—to the Ledger operator.
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+If the command fails, it prints an error message in place of that result, and the Job retries it up to
+three times (`backoffLimit: 3` in the manifest), starting a new Pod each time. If none of those
+attempts succeeds, the Job fails. Deploying the Job again without deleting the checkpoint volume
+resumes the processing from the same checkpoint data. To start the processing over from the beginning
+instead, see [Re-running after a failure](#re-running-after-a-failure).
+
+### Step 3 — Ledger AD: `cleanup-coordinator`
```
josh-wong:
```suggestion
### Step 3 – Ledger AD: `cleanup-coordinator`
```

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+**`cleanup-coordinator` (Ledger AD)**
+
+```bash
+kubectl -n "${LEDGER_NS:?}" delete job/scalardl-cleanup-coordinator
+: "${CLEANUP_VERSION:?}" "${LEDGER_TOKEN:?}" "${AUDITOR_TOKEN:?}" && envsubst < ledger-ad/cleanup-coordinator.yaml | kubectl -n "${LEDGER_NS:?}" apply -f -
+```
+
+On a resumed `cleanup-coordinator`, the token values are ignored — the deletion boundary saved in the
```
josh-wong:
```suggestion
On a resumed `cleanup-coordinator`, the token values are ignored—the deletion boundary saved in the
```

josh-wong (changes requested):
I've added some suggestions (mostly minor). PTAL!

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,376 @@
...
+  A2 -->|"Auditor token"| L3
+```
+
+## Limitations
+
+- The tool supports **Azure Cosmos DB for NoSQL** only. The tool rejects other configurations including `multi-storage`.
+- Ledger must not share the Coordinator table that its ScalarDB manages with any other ScalarDB
+  instance.
+- [`tx_state_management.enabled`](https://scalardl.scalar-labs.com/docs/latest/configurations/#tx_state_managementenabled) must have never been enabled.
+- [`coordinator.group_commit.enabled`](https://scalardb.scalar-labs.com/docs/latest/configurations/#coordinatorgroup_commitenabled) must have never been enabled.
```
jnmt, line 49: This might read as forbidding the Function use case, which is actually supported: when a function writes to a namespace in the Ledger's transaction, the ScalarDB Cluster that the application uses to access that same namespace has to share the Ledger's Coordinator table. Since `finalize-ledger` discovers and finalizes every transactional table in the storage it connects to, what actually matters seems to be **whether all the tables using that Coordinator table live in that storage**; so it would be worth stating that instead. Also, since we cannot validate another instance's configuration from this tool, the case we should call out explicitly is the shared instance using `multi-storage` (e.g., the ScalarDB Cluster side uses multi-storage, and the coordinator state table and a table for Function exist in Cosmos DB).

```suggestion
- The tool supports **Azure Cosmos DB for NoSQL** only. The tool rejects other configurations including `multi-storage`.
- Every transactional table that uses the Coordinator table that Ledger's ScalarDB manages must reside in the Cosmos DB account that the tool connects to.
- [`tx_state_management.enabled`](https://scalardl.scalar-labs.com/docs/latest/configurations/#tx_state_managementenabled) must have never been enabled.
- [`coordinator.group_commit.enabled`](https://scalardb.scalar-labs.com/docs/latest/configurations/#coordinatorgroup_commitenabled) must have never been enabled.

The second limitation still allows Ledger to share its Coordinator table with another ScalarDB instance, such as the ScalarDB Cluster that an application uses to access the namespaces that Ledger functions write to, because `finalize-ledger` finalizes every transactional table in the account no matter which instance owns it. What it rules out is such an instance (for example, ScalarDB Cluster) being configured with `multi-storage`: the tool cannot detect that configuration, and the instance's tables outside the account would keep their non-terminal records while the Coordinator state records for their transactions are deleted, leaving those records unrecoverable.
```

jnmt (commented):
Overall looks good, thank you! Left a comment on an important limitation. PTAL!

scalardl-cleanup/README.md
KodaiD: There can be cases where the tool is executed multiple times. So, I removed "once" in [5d9f943](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/5d9f943d40c9049eae38ff41059b4ba3d55fc073). Thanks!

scalardl-cleanup/README.md
KodaiD: Applied this in [5d9f943](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/5d9f943d40c9049eae38ff41059b4ba3d55fc073). Thanks!

scalardl-cleanup/README.md
KodaiD: Applied this in [5d9f943](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/5d9f943d40c9049eae38ff41059b4ba3d55fc073). Thanks!

scalardl-cleanup/README.md
KodaiD: Fixed in [0f759e4](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/0f759e49ad088ae9e626e483a4a8d0e238a91873). Thank you for letting me know!

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,323 @@
+# ScalarDL Cleanup
+
+## What is ScalarDL Cleanup?
```
feeblefakie: I think `ScalarDL Cleanup` is a bit odd as a tool name.
How about `ScalarDL Cleanup Tool`  or `ScalarDL Purger`?
For the latter, it might need to change the command name, though.

cc: @jnmt @josh-wong

feeblefakie (commented):
Overall, looking good. Left some comments. PTAL!

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+ScalarDL Cleanup removes the residual transaction state that a ScalarDL deployment accumulated while it was running a version of ScalarDL that does not support purge, which is a version earlier than ScalarDL 3.14.0. The residual transaction state consists of the following:
+
+- **Coordinator state records:** The transaction state records that ScalarDB manages in the Coordinator table on the Ledger side.
+- **Request proofs:** The records that Auditor stores for each client request to detect Byzantine faults.
+
+Because ScalarDL does not purge this state by default, it accumulates over time and consumes storage. ScalarDL 3.14.0 and later can purge it, but none of the purge options remove the state that an earlier version wrote, because that state lacks the information that purge requires. Use this tool to remove that state after you upgrade an existing deployment and before you enable purge on it. For details about the residual transaction state and the purge options, see [Purge the Residual Transaction State](https://scalardl.scalar-labs.com/docs/latest/purge-residual-transaction-state/).
```
kota2and3kan:
```suggestion
Because ScalarDL does not purge this state by default, it accumulates over time and consumes storage. ScalarDL 3.14.0 and later can purge it, but none of the purge options remove the state that an earlier version wrote, because that state lacks the information that purge requires. Use this tool to remove that state after you upgrade an existing deployment to 3.14.0 or later and before you enable purge on it. For details about the residual transaction state and the purge options, see [Purge the Residual Transaction State](https://scalardl.scalar-labs.com/docs/latest/purge-residual-transaction-state/).
```
This is a minor thing, but I think it would be better to explain the version (upgrade target) explicitly, just in case.

scalardl-cleanup/README.md
KodaiD: Good point. I've updated the description in [f10a6ca](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/f10a6ca7b96476fc3ec5e915182cc357f576882e) based on your suggestion. PTAL!

scalardl-cleanup/README.md
KodaiD: Applied this in [3abf381](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/3abf381d5fcb8256a951b4cde4d454d7332333df). Thanks!

scalardl-cleanup/README.md
KodaiD: I don't have a strong opinion on this, but `ScalarDL Purger` could be confused with manual purge?
What do you think? @jnmt @josh-wong @kota2and3kan

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,325 @@
+# ScalarDL Cleanup
+
+## What is ScalarDL Cleanup?
+
+ScalarDL Cleanup is a command-line tool that removes the residual transaction state that a ScalarDL deployment accumulated while it was running a version of ScalarDL that does not support purge, which is a version earlier than ScalarDL 3.14.0. The residual transaction state consists of the following:
+
+- **Coordinator state records:** The transaction state records that ScalarDB manages in the Coordinator table on the Ledger side.
+- **Request proofs:** The records that Auditor stores for each client request to detect Byzantine faults.
+
+Because ScalarDL does not purge this state by default, it accumulates over time and consumes storage. ScalarDL 3.14.0 and later can purge it, but none of the purge options remove the state that an earlier version wrote, because that state lacks the information that purge requires. Use this tool to remove that state after you upgrade an existing deployment and before you enable purge on it. For details about the residual transaction state and the purge options, see [Purge the Residual Transaction State](https://scalardl.scalar-labs.com/docs/latest/purge-residual-transaction-state/).
```
kota2and3kan: Just a question.

Do we need to follow the steps below strictly?

1. Upgrade to 3.14.0 or later.
2. Remove residual transaction states by using the cleanup tool.
3. Enable the purge feature.
4. Purge new transaction states that are generated by 3.14.0 or later.

In other words, can I do the following steps?

1. Upgrade to 3.14.0 or later.
2. Enable the purge feature.
3. Purge new transaction states that are generated by 3.14.0 or later.
4. Remove residual transaction states by using the cleanup tool.

For example, if the total amount of old version transaction states is huge, users might want to enable the server-side purge feature to avoid accumulating additional transaction states as a first step. And then, they might start removing old transaction states gradually by using the cleanup tool.

Therefore, I'd like to confirm whether the following description is a requirement or not.

```
Use this tool to remove that state after you upgrade an existing deployment and before you enable purge on it.
```

If it's not a requirement, I think it would be better to update this sentence.

Vice versa, if it's not a recommended way even if we technically can do it, I think we can keep the current explanation as is.

scalardl-cleanup/README.md
josh-wong: It's a little hard to come up with a descriptive name without sounding generic since the tool solves a specific issue.

`ScalarDL Residual State Cleanup`/`ScalarDL Residual State Purge` would seem to be the most accurate, but I think that might be a little too long. Shortening it to an acronym (like `ScalarDL RSC`/`ScalarDL RSP`) might be fine for internal use, but if we ever made it public, acronyms in general aren't very good for users since they require explaining what the acronym means (meaning that it makes understanding and adoption of the tool too high).

I'm not sure if removing `State` would help (like `ScalarDL Residual Cleanup`/`ScalarDL Residual Purge`), but that could be another option.

scalardl-cleanup/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+## How it works
+
+The tool provides three commands, split across the two administrative domains (ADs):
+
+| Command | AD | Description |
+|-----------------------|---------|--------------------------------------------------------------------------------------------------------------------------|
+| `finalize-ledger`     | Ledger  | Drives every unfinished record in every transactional table to a terminal state, then prints a **completion token**.    |
+| `finalize-auditor`    | Auditor | Releases every unreleased asset lock, deletes the request proofs that are safe to remove, then prints a **completion token**. |
+| `cleanup-coordinator` | Ledger  | Deletes the Coordinator state records that are safe to remove. Requires **both** completion tokens.                     |
```
kota2and3kan:
```suggestion
| `finalize-ledger`     | Ledger  | Drives every unfinished record in every transactional table to a terminal state, then prints a **completion token**. This command accesses Cosmos DB.                                |
| `finalize-auditor`    | Auditor | Releases every unreleased asset lock, deletes the request proofs that are safe to remove, then prints a **completion token**.  This command accesses Cosmos DB and ScalarDL Auditor. |
| `cleanup-coordinator` | Ledger  | Deletes the Coordinator state records that are safe to remove. Requires **both** completion tokens. This command accesses Cosmos DB                                                  |
```

I think it would be better to explain where each command accesses because it seems that there is no clear explanation of the access target in this document.

Also, I guess this explanation can tell users what configuration is necessary implicitly. It's helpful for users to understand this tool itself and its configurations for this tool.

What do you think?

kota2and3kan (commented):
Thank you for the updates!
I left several suggestions and questions.
Please take a look when you have time! 🙇

scalardl-cleanup/README.md
jnmt: I'm fine with `ScalarDL Cleanup Tool` (and `## What is the ScalarDL Cleanup Tool?`), which keeps the name consistent with the `scalardl-cleanup` image and CLI; those came out of an earlier naming discussion ("cleanup" is not the best, but it still has a noun usage, so it's fine or something like that). I would rather avoid `Purger`, as Kodai said.

jnmt (approved):
LGTM, thank you!

scalardl-cleanup/README.md
KodaiD: @josh-wong Thank you for your comment! How about `ScalarDL Cleanup Tool` that Hiro and Jun suggested? Is it more natural than `ScalarDL Cleanup` as a tool name?

scalardl-cleanup/README.md
KodaiD: Good point. Agreed with you. Applied this in [bdba611](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/bdba611f481d23476de08b37368c20715670b7a9). Thanks!

scalardl-cleanup/README.md
KodaiD: Good point. For now, the wording aligns with the documentation for the Purge feature. Fixed in bdba611. I'll update it once the description in the Purge feature doc is updated. Thanks!

scalardl-cleanup/README.md
KodaiD: Fixed in bdba611. Also added a limitation since it's important. Thanks!

scalardl-cleanup/README.md
josh-wong: Sure, I'm OK with `ScalarDL Cleanup Tool`.

We might need to consider later if we want to add `Tool` consistently in the naming for our other tools, like `Data Loader` and `Schema Loader`.

kota2and3kan (approved):
LGTM! Thank you!

scalardl-cleanup/README.md
KodaiD: Updated in [f60f88e](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/f60f88eb9ffa3319729e3f4763ca7ad758ceeaac) to refer to the purge doc.

scalardl-cleanup/README.md
KodaiD: Thank you for the confirmation! Renamed in [f60f88e](https://github.com/scalar-labs/scalardl-tools/pull/80/commits/f60f88eb9ffa3319729e3f4763ca7ad758ceeaac).

josh-wong (approved):
LGTM! Thank you!🙇🏻‍♂️

feeblefakie (approved):
LGTM! Thank you!

PR #79, opened 2026-09-02
Tell the operator to upgrade when the Auditor lacks RecoverAssetLock

KodaiD:
## Description

An Auditor without the `RecoverAssetLock` RPC makes `finalize-auditor` fail with gRPC's raw `UNIMPLEMENTED: Method not found: ...`, leaving the operator to work out that the fix is a server upgrade. This PR reports a user error that says so instead.

## Related issues and/or PRs

Depends on:

- #78 

## Changes made

- Add `RECOVER_ASSET_LOCK_RPC_UNSUPPORTED`, which names the missing RPC and tells the operator to upgrade the Auditor.
- Raise it from `LockFinalizer` when the exception chain carries gRPC's `UNIMPLEMENTED`, and let every other status propagate unchanged.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] I have considered whether similar issues could occur in other products, components, or modules if this PR is for bug fixes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [ ] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

N/A

## Release notes

N/A

PR #83, opened 2026-09-09
Add more E2E scenarios

KodaiD:
## Description

This PR takes the E2E test to the three cases the review of #55 asked for: a command interrupted midway, traffic arriving after the completion tokens are taken, and `finalize-auditor` run against a server too old for the RPC it needs.

Please focus your review on the added scenario sections.

## Related issues and/or PRs

- #55 
- #79 

## Changes made

- Add a scenario that runs `finalize-auditor` against an Auditor without `RecoverAssetLock` and asserts it fails as a user error naming the missing RPC.
- Add a scenario that interrupts each command mid-scan and asserts the replacement Pod resumes the same run rather than starting a new one.
- Run new traffic in the happy path once both tokens are taken, and assert the cleanup leaves that transaction state alone.
- Give every scenario its own fixture rebuild, so none of them depends on the order they run in.
- Assert the invariants directly in Cosmos: nothing left mid-transaction, no asset locks still held.
- Settle the comment conventions across `ci/e2e`.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,13 +1,6 @@
 name: E2E
 
-# End-to-end test for ScalarDL Cleanup. The tool reclaims garbage left behind by an older ScalarDL
-# server, while coordinating with the upgraded server. This test reproduces exactly that:
-#
-#   1. Deploy ScalarDL Ledger + Auditor at the old version on minikube against a real Cosmos DB,
-#      load the schema, and populate committed + stranded data with the old client (HashStore CLI).
-#   2. Upgrade both servers in place to the new version (which exposes the privileged
-#      RecoverAssetLock RPC that finalize-auditor needs).
-#   3. Run the scenario script in ci/e2e/scenarios.
+# End-to-end test for ScalarDL Cleanup against a real Cosmos DB. Runs scenarios in ci/e2e/scenarios.
 #
 # CI-only. Requires repository secrets: COSMOS_URI, COSMOS_KEY, CR_PAT.
 
@@ -107,6 +100,7 @@ jobs:
         run: minikube image load "ghcr.io/scalar-labs/scalardl-cleanup:$CLEANUP_VERSION"
 
       - name: Deploy ScalarDL (old version) + load schema
+        id: deploy
         env:
           COSMOS_URI: ${{ secrets.COSMOS_URI }}
           COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
@@ -114,30 +108,27 @@ jobs:
           SCALARDL_VERSION: ${{ env.SCALARDL_OLD_VERSION }}
         run: ./ci/e2e/manage-cluster.sh deploy
 
-      - name: Populate committed data
-        run: ./ci/e2e/populate.sh commit-objects
-
-      - name: Generate stranded locks (Ledger down)
-        run: ./ci/e2e/populate.sh strand-locks
-
-      - name: Write populated-assets metadata
-        run: ./ci/e2e/populate.sh write-metadata
-
-      - name: Upgrade ScalarDL to the new version
+      - name: Run scenario - Auditor without RecoverAssetLock
         env:
-          SCALARDL_VERSION: ${{ env.SCALARDL_NEW_VERSION }}
-        run: ./ci/e2e/manage-cluster.sh upgrade
+          COSMOS_URI: ${{ secrets.COSMOS_URI }}
+          COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
+          RECORD_COUNT: 1
+        run: ./ci/e2e/scenarios/old-auditor.sh
 
       - name: Run scenario - happy path
+        if: ${{ !cancelled() && steps.deploy.outcome == 'success' }}
+        env:
+          COSMOS_URI: ${{ secrets.COSMOS_URI }}
+          COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
         run: ./ci/e2e/scenarios/happy-path.sh
 
-      - name: Upload populated-assets
-        if: always()
-        uses: actions/upload-artifact@v7
-        with:
-          name: e2e-populated-assets
-          path: ${{ runner.temp }}/populated-assets.json
-          if-no-files-found: warn
+      - name: Run scenario - interrupt and resume
+        if: ${{ !cancelled() && steps.deploy.outcome == 'success' }}
+        env:
+          COSMOS_URI: ${{ secrets.COSMOS_URI }}
+          COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
+          RECORD_COUNT: 20 # Enough rows that the scan is still running when the Pod is deleted.
+        run: ./ci/e2e/scenarios/interrupt-resume.sh
 
       - name: Collect diagnostics on failure
         if: failure()
@@ -163,7 +154,6 @@ jobs:
         env:
           COSMOS_URI: ${{ secrets.COSMOS_URI }}
           COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
-          CR_PAT: ${{ secrets.CR_PAT }}
           SCALARDL_VERSION: ${{ env.SCALARDL_OLD_VERSION }}
         # Fatal on purpose: the shared Cosmos account must be left clean, so a failed
         # drop fails the job to alert an operator (who then drops it manually).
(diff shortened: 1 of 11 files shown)


scalardl-cleanup/ci/e2e/cleanup-jobs.sh
```diff
@@ -134,15 +245,38 @@ extract_completion_token() {
...
+  count_cosmos_records_by_query "$1" "$2" "$3" asset_lock \
+    'SELECT VALUE COUNT(1) FROM c WHERE c.values.lock_type IN (2, 3)'
+}
+
+# Count the records left mid-transaction: such a record still needs its Coordinator record, so zero
+# means cleanup-coordinator orphaned none. tx_state and its values are ScalarDB internals.
+# Usage: count_cosmos_unsettled_records <endpoint> <key> <database> <container>
+count_cosmos_unsettled_records() {
+  count_cosmos_records_by_query "$1" "$2" "$3" "$4" \
+    'SELECT VALUE COUNT(1) FROM c WHERE c.values.tx_state IN (1, 2)'
```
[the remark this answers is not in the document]
KodaiD: PREPARED = 1, DELETED = 2, COMMITTED = 3, ABORTED = 4, UNKNOWN = 5

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
```diff
@@ -3,88 +3,91 @@
...
+echo "request_proof rows: before=$rp_before_cleanup after=$rp_after_cleanup"
+[ "$rp_after_cleanup" -eq "$rp_before_cleanup" ] \
+  || { echo "::error::cleanup-coordinator changed request_proof from $rp_before_cleanup to $rp_after_cleanup"; exit 1; }
+
+# Nothing may still be pointing at the Coordinator records that were just deleted.
+for table in asset asset_metadata; do
+  unsettled=$(count_cosmos_unsettled_records "$ledger_uri" "$ledger_key" scalar "$table")
+  [ "$unsettled" -eq 0 ] \
+    || { echo "::error::$table holds $unsettled records left mid-transaction"; exit 1; }
+done
```
jnmt, line 86: Was it intentional that nothing in the fixture ever leaves a Ledger record in PREPARED/DELETED? `commit_objects()` commits cleanly and `strand_locks()` runs with the Ledger at 0 replicas, so this
count looks like it can only ever be 0. I understand it's complicated to create such a situation in Ledger, though.

scalardl-cleanup/ci/e2e/cleanup-jobs.sh
```diff
@@ -5,9 +5,10 @@
 #
 #   source ci/e2e/cleanup-jobs.sh
 #
-# Expects in the environment:
+# Required environment:
 #   CLEANUP_VERSION  tag of the scalardl-cleanup image
-#   COSMOSDB_SHELL   path to the Azure Cosmos DB Shell binary, used to count rows in Cosmos
+#   COSMOSDB_SHELL   path to the Azure Cosmos DB Shell binary (count_cosmos_records only, and it
+#                    checks for it itself)
```
jnmt: This PR adds `count_cosmos_held_locks` and `count_cosmos_unsettled_records`, which also need
`COSMOSDB_SHELL`; they all go through `count_cosmos_records_by_query`, which is what actually
checks for it, not `count_cosmos_records`.

scalardl-cleanup/ci/e2e/cleanup-jobs.sh
```diff
@@ -111,20 +132,110 @@ apply_job() {
...
+    (( SECONDS < deadline )) || diag_and_die "$ns" "job/$job did not start a Pod to resume with"
+    sleep 2
+  done
+  echo "job/$job resumed in pod/$resuming"
+
+  wait_for_job "$ns" "$job" "$timeout"
+  ok_pod=$(kubectl -n "$ns" get pod -l "job-name=$job" \
+    --field-selector=status.phase=Succeeded -o jsonpath='{.items[0].metadata.name}')
+  [ -n "$ok_pod" ] || diag_and_die "$ns" "no succeeded Pod found for job/$job in $ns"
+  kubectl -n "$ns" logs "$ok_pod" > "$RUNNER_TEMP/$job-resumed.log"
```
jnmt, line 237: Minor: `$resuming` is computed but never used — the assertions read whichever Pod ended up
Succeeded. With `backoffLimit: 3`, a resume that crashes once or twice before succeeding would
still pass, and `<job>-resumed.log` would hold the last Pod's log rather than the one that
resumed. Reading the log from `$resuming` instead would close both.

jnmt (commented):
Overall looks good, thank you! Left a few comments. PTAL 🙇‍♂️

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
KodaiD: @jnmt Good point. Thank you! Actually, the process of finalize-ledger is not verified in the test for now.
I'm planning to create a follow-up PR that tests finalize-ledger. My plan is:

- Create a ScalarDB table.
- Leave some prepared records with TwoPhaseCommitInterface's prepare().
- Check if finalize-ledger works as expected.

What do you think?

scalardl-cleanup/ci/e2e/cleanup-jobs.sh
KodaiD: Fixed in [7403843](https://github.com/scalar-labs/scalardl-tools/pull/83/commits/740384363772ebda2f4543aa5236a012dadc5116). Thanks!

scalardl-cleanup/ci/e2e/cleanup-jobs.sh
KodaiD: Good point. Rather than reading the log from $resuming, I asserted `[ "$ok_pod" = "$resuming" ]` in [7403843](https://github.com/scalar-labs/scalardl-tools/pull/83/commits/740384363772ebda2f4543aa5236a012dadc5116).  This closes both points and keeps the log file matching its name. Thanks!

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
jnmt: > Leave some prepared records with TwoPhaseCommitInterface's prepare().

Ahh, that's a good idea. Thanks for considering it.

jnmt (approved):
LGTM, thank you!

.github/workflows/e2e.yaml
```diff
@@ -107,37 +100,35 @@ jobs:
...
         env:
-          SCALARDL_VERSION: ${{ env.SCALARDL_NEW_VERSION }}
-        run: ./ci/e2e/manage-cluster.sh upgrade
+          COSMOS_URI: ${{ secrets.COSMOS_URI }}
+          COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
+          RECORD_COUNT: 1
+        run: ./ci/e2e/scenarios/old-auditor.sh
 
       - name: Run scenario - happy path
+        if: ${{ !cancelled() && steps.deploy.outcome == 'success' }}
```
brfrn169, line 119: With `!cancelled()` here, a failure in a scenario that isn't the last one gets overwritten: the next scenario's `setup-fixture.sh` deletes the cleanup Jobs with their Pods (`reset_checkpoint`) and restarts both servers, so `Collect diagnostics on failure` ends up showing the last scenario's cluster. That bites hardest on the new post-token check in happy-path — interrupt-resume has no counterpart to it, so it would pass and the diagnostics would describe a green run. (The run on 528a01a shows the mechanism: happy-path failed at 00:34, and the diagnostics at 00:40 list only interrupt-resume's Pods.)

Could we capture diagnostics per scenario — e.g. a diagnostics step keyed on `steps.<id>.outcome == 'failure'` after each one? Not a blocker; fine as a follow-up too.

brfrn169 (approved):
Left a minor comment. Other than that, LGTM! Thank you!

.github/workflows/e2e.yaml
KodaiD: Good point. I'll handle this in a separate PR. Thanks!

KodaiD: https://github.com/scalar-labs/scalardl-tools/pull/84

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
KodaiD: https://github.com/scalar-labs/scalardl-tools/pull/85

feeblefakie (approved):
LGTM! Thank you!

PR #85, opened 2026-09-17
Verify finalize-ledger settles non-terminal records in the E2E

KodaiD:
## Description

This PR makes the E2E actually exercise `finalize-ledger`. As pointed out in the review of #83, nothing in the fixture ever left a record in PREPARED or DELETED, so the non-terminal count the scenarios assert could only ever be 0 and the command's work went unverified. The fixture now creates such records deliberately with the ScalarDB Cluster SQL CLI, and the scenarios check that `finalize-ledger` settles them.

## Related issues and/or PRs

- #83 (this PR targets its branch)
- https://github.com/scalar-labs/scalardl-tools/pull/83#discussion_r3985465388

## Changes made

- Leave PREPARED and DELETED records in a namespace of the fixture's own, outside `scalar.asset` and `asset_metadata`.
- Deploy a standalone ScalarDB Cluster node sharing ScalarDL's Coordinator table, and run PREPARE without COMMIT against it.
- Assert in the happy-path and interrupt-resume scenarios that `finalize-ledger` settles every one of them, and that each record ends where aborting its transaction takes it.
- Skip CI when only the E2E scripts change to reduce the cost.

## Checklist

N/A

## Release notes

N/A

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,13 +1,6 @@
 name: E2E
 
-# End-to-end test for ScalarDL Cleanup. The tool reclaims garbage left behind by an older ScalarDL
-# server, while coordinating with the upgraded server. This test reproduces exactly that:
-#
-#   1. Deploy ScalarDL Ledger + Auditor at the old version on minikube against a real Cosmos DB,
-#      load the schema, and populate committed + stranded data with the old client (HashStore CLI).
-#   2. Upgrade both servers in place to the new version (which exposes the privileged
-#      RecoverAssetLock RPC that finalize-auditor needs).
-#   3. Run the scenario script in ci/e2e/scenarios.
+# End-to-end test for ScalarDL Cleanup against a real Cosmos DB. Runs scenarios in ci/e2e/scenarios.
 #
 # CI-only. Requires repository secrets: COSMOS_URI, COSMOS_KEY, CR_PAT.
 
@@ -107,6 +100,7 @@ jobs:
         run: minikube image load "ghcr.io/scalar-labs/scalardl-cleanup:$CLEANUP_VERSION"
 
       - name: Deploy ScalarDL (old version) + load schema
+        id: deploy
         env:
           COSMOS_URI: ${{ secrets.COSMOS_URI }}
           COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
@@ -114,30 +108,27 @@ jobs:
           SCALARDL_VERSION: ${{ env.SCALARDL_OLD_VERSION }}
         run: ./ci/e2e/manage-cluster.sh deploy
 
-      - name: Populate committed data
-        run: ./ci/e2e/populate.sh commit-objects
-
-      - name: Generate stranded locks (Ledger down)
-        run: ./ci/e2e/populate.sh strand-locks
-
-      - name: Write populated-assets metadata
-        run: ./ci/e2e/populate.sh write-metadata
-
-      - name: Upgrade ScalarDL to the new version
+      - name: Run scenario - Auditor without RecoverAssetLock
         env:
-          SCALARDL_VERSION: ${{ env.SCALARDL_NEW_VERSION }}
-        run: ./ci/e2e/manage-cluster.sh upgrade
+          COSMOS_URI: ${{ secrets.COSMOS_URI }}
+          COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
+          RECORD_COUNT: 1
+        run: ./ci/e2e/scenarios/old-auditor.sh
 
       - name: Run scenario - happy path
+        if: ${{ !cancelled() && steps.deploy.outcome == 'success' }}
+        env:
+          COSMOS_URI: ${{ secrets.COSMOS_URI }}
+          COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
         run: ./ci/e2e/scenarios/happy-path.sh
 
-      - name: Upload populated-assets
-        if: always()
-        uses: actions/upload-artifact@v7
-        with:
-          name: e2e-populated-assets
-          path: ${{ runner.temp }}/populated-assets.json
-          if-no-files-found: warn
+      - name: Run scenario - interrupt and resume
+        if: ${{ !cancelled() && steps.deploy.outcome == 'success' }}
+        env:
+          COSMOS_URI: ${{ secrets.COSMOS_URI }}
+          COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
+          RECORD_COUNT: 20 # Enough rows that the scan is still running when the Pod is deleted.
+        run: ./ci/e2e/scenarios/interrupt-resume.sh
 
       - name: Collect diagnostics on failure
         if: failure()
@@ -163,7 +154,6 @@ jobs:
         env:
           COSMOS_URI: ${{ secrets.COSMOS_URI }}
           COSMOS_KEY: ${{ secrets.COSMOS_KEY }}
-          CR_PAT: ${{ secrets.CR_PAT }}
           SCALARDL_VERSION: ${{ env.SCALARDL_OLD_VERSION }}
         # Fatal on purpose: the shared Cosmos account must be left clean, so a failed
         # drop fails the job to alert an operator (who then drops it manually).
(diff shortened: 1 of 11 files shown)


scalardl-cleanup/ci/e2e/db-records.sh
```diff
@@ -0,0 +1,196 @@
...
+}
+
+# Internal helper. Run a SQL script through the CLI and fail if any statement did. The CLI always
+# exits 0, so the output is what has to be judged.
+# Usage: db_records_sql <script-file>
+db_records_sql() {
+  local name log
+  name="$(basename "$1" .sql)"
+  log="$RUNNER_TEMP/$name.log"
+  java -jar "$SCALARDB_SQL_CLI" --config "$db_records_config" --file "$1" > "$log" 2>&1 || true
```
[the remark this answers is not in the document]
KodaiD: Fixed in [45e454e](https://github.com/scalar-labs/scalardl-tools/pull/85/commits/45e454e9ba6df772dcbfba04fcbd6718ab0315b8).

.github/workflows/ci.yaml
```diff
@@ -5,10 +5,12 @@ on:
     branches: [master]
```
KodaiD, line 5: Let's merge #83 first.

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
```diff
@@ -79,12 +94,13 @@ echo "request_proof rows: before=$rp_before_cleanup after=$rp_after_cleanup"
...
 # Nothing may still be pointing at the Coordinator records that were just deleted.
-for table in asset asset_metadata; do
-  unsettled=$(count_cosmos_unsettled_records "$ledger_uri" "$ledger_key" scalar "$table")
-  [ "$unsettled" -eq 0 ] \
-    || { echo "::error::$table holds $unsettled records left mid-transaction"; exit 1; }
-done
-echo "no records left mid-transaction in asset or asset_metadata"
+assert_no_unsettled_records "$ledger_uri" "$ledger_key" "$DB_NS" "$DB_TABLE"
+assert_no_unsettled_records "$ledger_uri" "$ledger_key" scalar asset
+assert_no_unsettled_records "$ledger_uri" "$ledger_key" scalar asset_metadata
```
jnmt, line 100: Nitpick: The two `scalar` ones now read like the `e2e_db.records` check above, but they can still only ever be 0. Fine as is, but could we note that they're only a sanity check? Same in `interrupt-resume.sh`.

E.g., "The scalar ones are only a sanity check: setup-fixture.sh leaves no non-terminal record in the Ledger's own tables.

.github/workflows/e2e.yaml
```diff
@@ -41,6 +41,8 @@ jobs:
       # than "latest" keeps the default imagePullPolicy at IfNotPresent, so the locally loaded image
       # is used and nothing is pulled from ghcr.
       CLEANUP_VERSION: dev
+      # The ScalarDB the cleanup tool builds against; the fixture's node and SQL CLI track it.
+      SCALARDB_VERSION: 3.19.1
```
jnmt: The comment above says the fixture tracks the ScalarDB the tool builds against, but nothing
enforces it — a bump to `scalarDbVersion` in `build.gradle` alone would silently leave the node and
the SQL CLI on the old version. Could we derive it instead, the way `ci.yml` in scalardl-enterprise
reads `scalarDlVersion` with `./gradlew :auditor:properties`?

jnmt (commented):
Overall looks good, thank you! Left two comments. PTAL!

scalardl-cleanup/ci/e2e/scenarios/happy-path.sh
KodaiD: I've added a comment in [5c3132b](https://github.com/scalar-labs/scalardl-tools/pull/85/commits/5c3132bbd20576143494ad469ce9d36f34dbad1d). Thanks!

.github/workflows/e2e.yaml
KodaiD: Good point. Fixed in [5c3132b](https://github.com/scalar-labs/scalardl-tools/pull/85/commits/5c3132bbd20576143494ad469ce9d36f34dbad1d). Thanks!

jnmt (approved):
LGTM, thank you!

scalardl-cleanup/ci/e2e/scenarios/interrupt-resume.sh
```diff
@@ -0,0 +1,182 @@
...
+
+setup_ledger_ad
+interrupt_and_resume "$LEDGER_NS" scalardl-finalize-ledger \
+  "$work/ledger-ad/finalize-ledger.yaml" "$SCAN_STARTED"
+assert_resumed_same_run scalardl-finalize-ledger
+ledger_out=$(read_job_output_json "$LEDGER_NS" scalardl-finalize-ledger)
+ledger_token=$(extract_completion_token finalize-ledger "$ledger_out")
+
+# Between the interrupted run and the resumed one, every record has to end where an uninterrupted
+# run would leave it -- none skipped, none finalized twice.
```
brfrn169, line 115: In the green run on 5c3132b, the interruption never reaches `e2e_db.records`, and it can't as things stand. finalize-ledger works through its tables one at a time, in the order of the `HashSet`s `CosmosAdmin` builds from the namespace and table names. With these names that order is fixed: `scalar.asset_metadata` → `scalar.asset` → `e2e_db.records`. So the first `Cosmos DB physical partitions:` line, which is what the kill waits for, always belongs to `scalar.asset_metadata`. The log matches: the resumed Pod reports `Scan complete` for all three tables (`scalar.asset_metadata: 14 records`, `scalar.asset: 20 records`, `e2e_db.records: 60 records`), so none had finished before the kill, and all 60 `e2e_db.records` rows were read by the resumed Pod in one pass.

The interrupted Pod therefore never meets a non-terminal record. The checks below end up confirming what happy-path already does, and "none skipped, none finalized twice" isn't exercised.

Could we have finalize-ledger's kill wait for `FeedRanges for e2e_db.records` instead (the scanner logs `Discovered N FeedRanges for <table>` as it starts each table), and check that the resumed log has `Loaded N persisted FeedRanges for e2e_db.records`, so a kill that lands too late fails instead of passing? If that's more than this PR wants, could we reword this comment and the one at L101 so they don't claim the interruption covers these records?

scalardl-cleanup/ci/e2e/db-records.sh
```diff
@@ -0,0 +1,203 @@
...
+E2E_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+source "$E2E_DIR/cleanup-jobs.sh"
+source "$E2E_DIR/port-forward.sh"
+
+# Exposed because a scenario counts rows in Cosmos itself.
+DB_NS="e2e_db"
+DB_TABLE="records"
+
+# Mirrors RecoveryHandler.n in ScalarDB: how long recovery leaves a prepared record alone.
```
brfrn169, line 26: `RecoveryHandler.n` doesn't exist in ScalarDB; the constant is `RecoveryHandler.TRANSACTION_LIFETIME_MILLIS` (15000 in v3.19.1, checked against `tx_prepared_at` in `isTransactionExpired()`). Since the ScalarDB version now follows the build, this is the one value someone has to recheck by hand on a bump, so it helps if the pointer is greppable, like `LockOrderRecoveryHandler.LOCK_VALID_PERIOD_MILLIS` in populate.sh.
```suggestion
# Mirrors RecoveryHandler.TRANSACTION_LIFETIME_MILLIS in ScalarDB: how long recovery leaves a
# prepared record alone.
```

brfrn169 (approved):
Left a few minor comments. Other than that, LGTM! Thank you!

scalardl-cleanup/ci/e2e/scenarios/interrupt-resume.sh
KodaiD: Fixed in [1d4aec8](https://github.com/scalar-labs/scalardl-tools/pull/85/commits/1d4aec8e60c737b8f151bb59a342f5cac0c725e9). Thanks!

scalardl-cleanup/ci/e2e/db-records.sh
KodaiD: Fixed in [1d4aec8](https://github.com/scalar-labs/scalardl-tools/pull/85/commits/1d4aec8e60c737b8f151bb59a342f5cac0c725e9). Thanks!

